PR #5, opened 2015-07-30
[WIP] More tests

dignifiedquire:
Very slowly I'm getting to know the code base. Hopefully more interesting tests coming soon.

### DIFF
--- a/src/swarm.js
+++ b/src/swarm.js
@@ -150,9 +150,9 @@ function Swarm () {
     if (number === 0) { cb() }
     var c = new Counter(number, cb)
 
-    keys.map(function (key) {
-      c.hit()
+    keys.forEach(function (key) {
       self.connections[key].conn.end()
+      c.hit()
     })
   }
 
--- a/src/swarm.js
+++ b/src/swarm.js
@@ -20,7 +20,7 @@ function Swarm () {
 
   self.port = parseInt(process.env.IPFS_SWARM_PORT, 10) || 4001
   self.connections = {} // {peerIdB58: {conn: <>, socket: <>}
-  self.handles = []
+  self.handles = {}
 
   // set the listener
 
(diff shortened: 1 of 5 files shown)


daviddias:
:+1: Awesome! Thank you @Dignifiedquire , want me to merge it now or want to add more tests to this PR?

dignifiedquire:
@diasdavid found my first bug while adding tests :) 
If you want you can merge it, and I'll just do new PRs with more tests

dignifiedquire:
Needs https://github.com/diasdavid/node-spdy-stream-muxer/pull/3 to pass the tests

src/swarm.js
```diff
@@ -20,7 +20,7 @@ function Swarm () {
 
   self.port = parseInt(process.env.IPFS_SWARM_PORT, 10) || 4001
   self.connections = {} // {peerIdB58: {conn: <>, socket: <>}
-  self.handles = []
+  self.handles = {}
```
daviddias, line 23: Any specific reason on having handles as an object instead of an array? I see you changed the looped and both work fine. I might just be missing a perf improv, perhaps?

daviddias:
Thank you @Dignifiedquire, this looking very good! I've went ahead and merged this PR, so that you don't have to be "linking" the patched version of spdy-stream-muxer.

Plus is awesome to have bugs fixes pushed :) That was ⭐️

src/swarm.js
dignifiedquire, line 23: No perf reason but the lookup you were trying to do if there was alread a handler wasn't working, so I changed it to an object, as otherwise we would have to search the whole list to check if we already have a handler.

dignifiedquire:
Thanks happy to help, even though I still don't understand all the dark connection magic. 

Btw you need to turn on travis, otherwise that badge will stay gray ;)

PR #7, opened 2015-08-02
identify: Fix some issues with updateSelf.

dignifiedquire:


tests/swarm-test.js
```diff
@@ -38,9 +38,16 @@ beforeEach(function (done) {
 })
 
 afterEach(function (done) {
-  swarmA.closeListener()
-  swarmB.closeListener()
-  done()
+  // This should be 2, but for some reason
+  // that will fail in most of the tests
```
dignifiedquire, line 42: @diasdavid not sure what is going wrong here, but we need to wait for all things to be closed to start the next test, otherwise I get race conditions and ports in use errors.

src/identify/index.js
```diff
@@ -118,8 +118,8 @@ function updateSelf (peerSelf, observedAddr) {
   }
 
   for (var i = 0; i < peerSelf.previousObservedAddrs.length; i++) {
-    if (peerSelf.previousObservedAddrs[i].toString() === omh.toString()) {
-      peerSelf.previousObserveredAddrs.splice(i, 1)
+    if (multiaddr(peerSelf.previousObservedAddrs[i]).toString() === omh.toString()) {
```
daviddias: Could you clarify this change? the `previousObserverAddrs is intentionally and array of`multiaddr`, they should not be stored as an array of strings.

tests/swarm-test.js
daviddias, line 42: I see. When one of the swarm closes, the other swarm will get a ECONRESET on the socket and close itself. That should why the second `closeListener` is failing and not properly returning. 

Not sure why it is failing down. A ECONRESET on one of the "listened" should not crash the main socket listen. Need to dig deeper in this one. Thank you @Dignifiedquire

src/identify/index.js
dignifiedquire: I didn't change what is stored in `previousObservedAddrs`. The issue was that when inspecting `observedAddr` it is a binary buffer and not a string buffer, so this equality check would always fail. So maybe https://github.com/Dignifiedquire/node-ipfs-swarm/blob/identify-tests/src/identify/index.js#L128 should be changed to 

``` js
peerSelf.previousObservedAddrs.push(omh)
```

daviddias: Good catch and yes, that is indeed the solution, that was my intent but it seems that I missed it. Thank you again :) 

Apologies for the delay in my response, I've been traveling for the last day and half to Battlemesh.

daviddias: Going to update that on master, so then that we can solve the ECONRESET issue in this PR with the tests you made

dignifiedquire:
Rebased onto the latest master to include the fixes made there

daviddias:
Going to merge this one and figure it out while we factor swarm. Going to separate the transports and the dialers with clear interfaces so we can add more than tcp

PR #11, opened 2015-09-25
Protect from peers without supported transports

masylum:
# What

Trying to run compliance tests from the kad router module. I've tried to port the new swarm API but forgot to add a transport. The tests ended up blowing up instead of failing gracefully.
# How to test

``` js
peerOne = new Peer(Id.create(), [multiaddr('/ip4/127.0.0.1/tcp/8090')])
peerTwo = new Peer(Id.create(), [multiaddr('/ip4/127.0.0.1/tcp/8091')])
swarm = new Swarm(peerOne)
swarm.dial(peerTwo, {}, function (err) {
  console.log(err);
});
```

This just work and display the error.

### DIFF
--- a/src/swarm.js
+++ b/src/swarm.js
@@ -126,6 +126,11 @@ function Swarm (peerInfo) {
       })
     })
 
+    if (!multiaddrs.length) {
+      callback(new Error("The swarm doesn't support any of the peer transports"))
+      return
+    }
+
     var conn
 
     async.eachSeries(multiaddrs, function (multiaddr, next) {
@@ -175,6 +180,7 @@ function Swarm (peerInfo) {
       // if protocol is selected, multistream that protocol
       if (!conn) {
         callback(new Error('Unable to open a connection'))
+        return
       }
 
       if (self.muxers['spdy']) {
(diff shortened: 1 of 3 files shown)


src/swarm.js
```diff
@@ -175,6 +180,7 @@ function Swarm (peerInfo) {
       // if protocol is selected, multistream that protocol
       if (!conn) {
         callback(new Error('Unable to open a connection'))
+        return
```
masylum, line 183: I couldn't easily find an scenario to test when this can happen... but without this `return` you can end up with double callbacks :scream:

tests/swarm-test.js
```diff
@@ -13,17 +13,35 @@ var Swarm = require('../src')
...
 })
 
+experiment('Basics', function () {
+  test('enforces creation with new', function (done) {
+    expect(function () {
+      Swarm()
+    }).to.throw()
+    done()
+  })
+})
```
masylum, line 28: couldn't help myself...

daviddias:
Nice catch! :D 

(for some weird reason, github didn't show this two PR you did on swarm in the last couple of days, sorry for have kept you waiting!)

PR #12, opened 2015-09-25
Do not force a callback in addTransport

masylum:
# What

Make callback optional
# How to test

``` js
swarmZero.addTransport('tcp', tcp, { multiaddr: mh }, {}, {port: 8090})
```

Should work

### DIFF
--- a/src/swarm.js
+++ b/src/swarm.js
@@ -63,7 +63,9 @@ function Swarm (peerInfo) {
         self.peerInfo.multiaddrs.push(options.multiaddr)
       }
 
-      callback()
+      if (callback) {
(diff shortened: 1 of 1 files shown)


jbenet:
LGTM. @diasdavid ?

daviddias:
Ok, can you add to the Docs that not adding the callback means that simply calling the function doesn't translate to automatically start listening? specially for transports like libp2p-webrtc, the set up time might be even longer, which can lead to racing condition bugs

masylum:
Good point @diasdavid, in that sense I think that the API is better as it is.

PR #13, opened 2015-09-25
Do not allow undefined `peerInfo`

masylum:
# What

The code assumes that `peerInfo` exists, the API doesn't.
# How to test

``` js
swarm = new Swarm()
swarm.addTransport('tcp', tcp, { multiaddr: mh }, {}, {port: 8095})
```

This shouldn't explode because of `self.peerInfo.multiaddrs`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,3 @@
 var Swarm = require('./swarm')
 
 exports = module.exports = Swarm
-exports.singleton = new Swarm()
--- a/tests/swarm-test.js
+++ b/tests/swarm-test.js
@@ -24,6 +24,16 @@ process.on('uncaughtException', function (err) {
   console.log('Caught exception: ' + err)
 })
 
+experiment('Without a peer', function () {
+  test('it throws an exception', function (done) {
+    expect(function () {
+      var sw = new Swarm()
+      sw.close()
+    }).to.throw(Error)
+    done()
+  })
+})
+
 experiment('Without a Stream Muxer', function () {
   experiment('tcp', function () {
     test('add the transport', function (done) {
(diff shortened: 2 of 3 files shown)


jbenet:
LGTM @diasdavid ?

daviddias:
Sweet! Thank you :) Can you add a section to tests like

``` JavaScript
experiment('Basics', function () {
 // all the tests
})
```

So that we increase our code coverage :)

src/index.js
```diff
@@ -1,4 +1,3 @@
 var Swarm = require('./swarm')
 
 exports = module.exports = Swarm
-exports.singleton = new Swarm()
```
masylum, line 4: not sure if this is used anywhere but better not since that singleton `Swarm` could not add transports.

tests/swarm-test.js
```diff
@@ -24,6 +24,16 @@ process.on('uncaughtException', function (err) {
   console.log('Caught exception: ' + err)
 })
 
+experiment('Without a peer', function () {
+  test('it throws an exception', function (done) {
+    expect(function () {
+      var sw = new Swarm()
+      sw.close()
```
masylum, line 31: had to do something with it because annoying `standard` was complaining on "Do not use 'new' for side effects"

src/index.js
daviddias, line 4: It was a old idea, where instead of passing the object around for Peer Routing, Distributed Record Store and Discovery mechanisms, one would simply do require('swarm').singleton. But eventually I never used it and just passed the swarm object along for making it more explicit. Good call on removing it! :)

daviddias:
LGTM :D 

Thank you @masylum ! :)

PR #14, opened 2015-09-26
Implemented `close` and improved the tests with it

masylum:
#### :tophat: What? Why?

I implemented the `close` method. I took the chance to improve the test suite to use it instead of keeping counters everywhere.

That being said I stumbled in the SPDY bug where connections are not closed entirely. Therefore I couldn't use `close` to refator those.
#### :ghost: GIF

^ me with the SPDY tests

### DIFF
--- a/src/swarm.js
+++ b/src/swarm.js
@@ -47,6 +47,12 @@ function Swarm (peerInfo) {
     // set up the transport and add the list of incoming streams
     // add transport to the list of transports
 
+    var multiaddr = options.multiaddr
+    if (multiaddr) {
+      // no need to pass that to the transports
+      delete options.multiaddr
+    }
+
     var listener = transport.createListener(options, listen)
 
     listener.listen(listenOptions, function ready () {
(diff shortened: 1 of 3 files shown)


src/swarm.js
```diff
@@ -47,6 +47,12 @@ function Swarm (peerInfo) {
     // set up the transport and add the list of incoming streams
     // add transport to the list of transports
 
+    var multiaddr = options.multiaddr
+    if (multiaddr) {
+      // no need to pass that to the transports
+      delete options.multiaddr
+    }
+
```
masylum, line 55: couldn't resist to do this one...

tests/swarm-test.js
```diff
@@ -25,164 +28,89 @@ process.on('uncaughtException', function (err) {
 })
 
 experiment('Without a Stream Muxer', function () {
-  experiment('tcp', function () {
+  experiment('and one swarm over tcp', function () {
```
daviddias, line 31: Do we need this degree of separation between 'one swarm' and 'two swarms'. 'one swarm' is just to test if the construction doesn't throw.

daviddias, line 31: Ah got it, it was so that you could have the 'beforeEach' and 'afterEach'

Nice! :)

src/swarm.js
daviddias, line 55: ok :) (although in the future, things like libp2p-webrtc might like that multiaddr, let's see, I might bring it back)

daviddias:
This is really great @masylum ! Thank you :D

Issue #22, opened 2016-03-10
add libp2p-websockets to the set of tests (plus browser test it)


diasdavid:

diasdavid:
just missing now `high level API - 1st **with** stream multiplexing (on websockets)`

PR #29, opened 2016-03-16
Browser tests

fbaiodias:
related to #22


daviddias:
@xicombd could you push the things we worked to the branch?

daviddias:
Actually, if possible, can you do the websockets tests between browser - node in a separate PR (using the .transport api and the .dial api), deferring the ones with spdy and identify on top

PR #30, opened 2016-03-20
Dial fixes

fbaiodias:
- Update libp2p-websockets and add test for callback's conn
- Return conn on swarm.dial


daviddias:
woot! :)

PR #34, opened 2016-03-23
Use dignified.js

dignifiedquire:
https://github.com/dignifiedquire/dignified.js/issues/1


daviddias:
thank you @dignifiedquire :D

PR #41, opened 2016-04-20
feat: handle ipfs multiaddrs

### LINKED ISSUE #38: Support for multiaddrs with ipfs
Following https://github.com/jbenet/js-multiaddr/pull/18/files. Now that multiaddr supports ipfs addrs. We should update swarm tests to use ipfs multiaddrs as well, this means that: swarm has to understand how to decapsulate the ipfs addr part when giving the multiaddr to the transport (the transport doesn't have to know about ipfs addr if it doesn't need)

- transport dial https://github.com/diasdavid/js-libp2p-swarm/blob/master/src/index.js#L70
- multiaddrs to be complemented with ipfs part https://github.com/diasdavid/js-libp2p-swarm/blob/master/test/08-swarm-without-muxing.node.js#L23-L24

dignifiedquire:
Closes #38

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,25 +40,26 @@
     "bl": "^1.1.2",
     "buffer-loader": "0.0.1",
     "chai": "^3.5.0",
-    "aegir": "^2.1.0",
+    "aegir": "^2.1.1",
     "gulp": "^3.9.1",
-    "istanbul": "^0.4.2",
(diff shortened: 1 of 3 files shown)


daviddias:
woot! Nice

@xicombd, you won't hit this issue anymore :)

PR #42, opened 2016-04-24
fix(identify): convert all addresses to multiaddr

### LINKED ISSUE #37: Uncaught TypeError: Argument must be a string
Only a few times, one of the tests fails with the following error:

```
 1) high level API - with everything mixed all together! dial from tcp+ws to tcp+ws:
     Uncaught TypeError: Argument must be a string
      at TypeError (native)
      at Buffer.write (buffer.js:580:21)
      at Object.exports.bytes.encode (node_modules/protocol-buffers/encodings.js:35:17)
      at Object.encode (eval at <anonymous> (node_modules/generate-function/index.js:55:21), <anonymous>:26:14)
      at Messages._push (node_modules/protocol-buffers-stream/index.js:73:9)
      at Messages.(anonymous function) [as identify] (node_modules/protocol-buffers-stream/index.js:50:12)
      at null.<anonymous> (src/identify.js:90:11)
      at Messages._emit (node_modules/protocol-buffers-stream/index.js:121:40)
      at Messages._onmessage (node_modules/protocol-buffers-stream/index.js:141:14)
      at Messages._write (node_modules/protocol-buffers-stream/index.js:94:38)
      at doWrite (node_modules/readable-stream/lib/_stream_writable.js:313:64)
```

This happens inside the identify protocol, when we get back the identify protobuf and protocol-buffer-stream tried to parse. I've tried to diagnose this issue and was unable to find the root cause (specially because it only happens few times, and when I think it is solved, comes back after to torment me again).

dignifiedquire:
Fixes #37

In addition https://github.com/diasdavid/js-peer-info/pull/16 makes this a lot safer

### DIFF
--- a/src/identify.js
+++ b/src/identify.js
@@ -41,7 +41,7 @@ exports.exec = (rawConn, muxer, peerInfo, callback) => {
 
       pbs.on('identify', (msg) => {
         if (msg.observedAddr.length > 0) {
-          peerInfo.multiaddr.addSafe(msg.observedAddr)
+          peerInfo.multiaddr.addSafe(multiaddr(msg.observedAddr))
         }
 
         const peerId = Id.createFromPubKey(msg.publicKey)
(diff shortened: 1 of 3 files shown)


daviddias:
LGTM :) Thank you for solving this so annoying bug!

PR #44, opened 2016-04-26
feature/events

daviddias:
Adding necessary events so that bitswap can act on new peers or peers leaving

//cc @dignifiedquire

### DIFF
--- a/package.json
+++ b/package.json
@@ -56,7 +56,6 @@
   "dependencies": {
     "babel-runtime": "^6.6.1",
     "duplex-passthrough": "github:diasdavid/duplex-passthrough",
-    "eventemitter2": "^1.0.3",
     "ip-address": "^5.8.0",
     "lodash.contains": "^2.4.3",
     "multistream-select": "^0.6.5",
--- a/package.json
+++ b/package.json
@@ -56,6 +56,7 @@
   "dependencies": {
     "babel-runtime": "^6.6.1",
     "duplex-passthrough": "github:diasdavid/duplex-passthrough",
+    "eventemitter2": "^1.0.3",
     "ip-address": "^5.8.0",
     "lodash.contains": "^2.4.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -56,6 +56,7 @@
   "dependencies": {
     "babel-runtime": "^6.6.1",
     "duplex-passthrough": "github:diasdavid/duplex-passthrough",
+    "eventemitter2": "^1.0.3",
```
dignifiedquire: Why this eventemitter and not the default one from node/webpack/browserify?

src/index.js
```diff
@@ -157,6 +161,13 @@ function Swarm (peerInfo) {
           this.muxedConns[pi.id.toB58String()] = {}
           this.muxedConns[pi.id.toB58String()].muxer = muxedConn
           this.muxedConns[pi.id.toB58String()].conn = conn // to be able to extract addrs
+
+          self.emit('peer-mux-established', pi)
+
+          muxedConn.on('close', () => {
+            delete self.muxedConns[pi.id.toB58String()]
+            self.emit('peer-mux-closed', pi)
```
dignifiedquire, line 169: Could we just call it `peer-disconnect`, as that feels like it would be easier to understand (and to remember)

package.json
daviddias: I've been using this one for a while(years) now. Before this was the one that ran 100% on the browser. But I can with all the transpilation nowadays, we can use the default one.

src/index.js
daviddias, line 169: We could. The thing I'm trying to distinguish, is that there is 'conns' and 'muxedConns' and for bitswap, it only really matters the 'muxedConns'. Saying 'peer-disconnect' and not emitting it for a regular conn exit could mislead a user.

daviddias, line 169: Would you like to propose a better name?

PR #45, opened 2016-05-04
attach peerId to the conn

daviddias:
This gets what we wanted/need.

Not super happy with the fact that since identify to finish (and since this is all async), there are some incoming connections that are delivered to their protocol handlers without that connId (typically just the first one, during the period where identify is still doing its thing).


dignifiedquire:
lgtm

PR #64, opened 2016-05-22
restructure and add spdy to browser tests

daviddias:
> **WIP**


gulpfile.js
```diff
@@ -31,16 +35,21 @@ gulp.task('test:browser:before', (done) => {
     swarmB = new Swarm(peerB)
 
     swarmB.transport.add('ws', new WebSockets())
-    swarmB.transport.listen('ws', {}, null, cb)
+    swarmB.transport.listen('ws', {}, null, () => {
+      // swarmB.connection.addStreamMuxer(spdy)
+      // swarmB.connection.reuse()
```
daviddias, line 40: @dignifiedquire spdy (or identify) wasn't being used for browser tests and I found that if I run them:
- with spdy - they pass locally, fail on CI
- with identify - they fail

However, this is combination is used in several other places, including orbit. Are these tests missing anything?

test/browser-03-swarm-with-muxing-plus-webrtc-star.js
```diff
@@ -0,0 +1,94 @@
...
+
+  it('listen on swarm 2', (done) => {
+    swarm2.listen(done)
+  })
+
+  it('add spdy', () => {
+    swarm1.connection.addStreamMuxer(spdy)
+    // swarm1.connection.reuse()
+    swarm2.connection.addStreamMuxer(spdy)
+    // swarm2.connection.reuse()
```
daviddias, line 61: Also not being successful with Identify on top of webrtc-star (in the browser)

I'm really intrigued, since identify has been a key enabler for Orbit and there are so many successful tests.

gulpfile.js
dignifiedquire, line 40: Are you seeing these issues with `tcp` as well?

daviddias, line 40: nop. It doesn't even happen with websockets just in Node.js, it only happens in the browser tests

daviddias:
Going to merge this and update libp2p-ipfs-browser, so that then we can test js-ipfs with bitswap on top of it :)

PR #68, opened 2016-05-24
Make identify not freak out on missing pubkey, it is ok.. this enables webrtc-star discovery to work fine

daviddias:


dignifiedquire:
LGTM

PR #72, opened 2016-05-30
fix: Uncaught Error: socket hang up

daviddias:
WIP

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   "dependencies": {
     "babel-runtime": "^6.6.1",
     "browserify-zlib": "github:ipfs/browserify-zlib",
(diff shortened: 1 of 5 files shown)


daviddias:
I've just pushed my latest changes (for debugging) onto this branch, to test it quickly, run with `./node_modules/.bin/mocha test/05-muxing-spdy.node.js`

daviddias:
I've also added listeners for the error and close events on the tcp sockets created (inside the libp2p-tcp that gets installed on the node_modules) just to check if the error was leaking. I indeed catch the error on the 'close' event, as expected, but still, it gets thrown as an uncaught exception

dignifiedquire:
lgtm with a better commit message

daviddias:
better commit message added. This is a great bug fix! :D

PR #77, opened 2016-06-10
[WIP] Make identify go compatible

dignifiedquire:
Current state
- Dialing to go fails because go does not send a public key
- Dialing from go to js just hangs it seems no data arrives on the conn pass to `identify.handle`


dignifiedquire:
cc @diasdavid @whyrusleeping

daviddias:
Closing in favour of https://github.com/libp2p/js-libp2p-swarm/pull/80

PR #79, opened 2016-06-23
update the transports

daviddias:
This PR updates the transport dependencies and adapts swarm to work with the interface-transport and interface-connection, without compromising any guarantees or interface.

### DIFF
--- a/src/dial.js
+++ b/src/dial.js
@@ -1,9 +1,9 @@
 'use strict'
 
 const multistream = require('multistream-select')
-const Duplexify = require('duplexify')
+const Connection = require('interface-connection').Connection
 
-const connHandler = require('./default-handler')
+const protocolMuxer = require('./protocol-muxer')
 
 module.exports = function dial (swarm) {
   return (pi, protocol, callback) => {
@@ -16,7 +16,7 @@ module.exports = function dial (swarm) {
       callback = function noop () {}
     }
 
-    const pt = new Duplexify()
+    const proxyConn = new Connection()
 
     const b58Id = pi.id.toB58String()
 
@@ -40,7 +40,7 @@ module.exports = function dial (swarm) {
       gotMuxer(swarm.muxedConns[b58Id].muxer)
     }
 
-    return pt
+    return proxyConn
 
     function gotWarmedUpConn (conn) {
       attemptMuxerUpgrade(conn, (err, muxer) => {
@@ -145,7 +145,7 @@ module.exports = function dial (swarm) {
             // in case identify is on
             muxedConn.on('stream', (conn) => {
               conn.peerId = pi.id
-              connHandler(swarm.protocols, conn)
+              protocolMuxer(swarm.protocols, conn)
             })
 
             cb(null, muxedConn)
@@ -168,11 +168,9 @@ module.exports = function dial (swarm) {
           if (err) {
             return callback(err)
           }
-
-          pt.setReadable(conn)
-          pt.setWritable(conn)
-          pt.peerId = pi.id
-          callback(null, pt)
+          proxyConn.setInnerConn(conn)
+          proxyConn.peerId = pi.id
+          callback(null, proxyConn)
         })
       })
     }
--- a/src/identify.js
+++ b/src/identify.js
@@ -54,7 +54,7 @@ exports.exec = (rawConn, muxer, pInfo, callback) => {
             return callback(err)
           }
           const msg = idPb.Identify.decode(data)
-          if (msg.observedAddr.length > 0) {
+          if (hasObservedAddr(msg)) {
             pInfo.multiaddr.addSafe(multiaddr(msg.observedAddr))
           }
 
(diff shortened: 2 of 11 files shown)


src/dial.js
```diff
@@ -168,11 +168,9 @@ module.exports = function dial (swarm) {
           if (err) {
             return callback(err)
           }
-
-          pt.setReadable(conn)
-          pt.setWritable(conn)
-          pt.peerId = pi.id
-          callback(null, pt)
+          proxyConn.setInnerConn(conn)
+          proxyConn.peerId = pi.id
```
dignifiedquire, line 172: Is this using a setter method, or direct property access?

src/dial.js
```diff
@@ -168,11 +168,9 @@ module.exports = function dial (swarm) {
           if (err) {
             return callback(err)
           }
-
-          pt.setReadable(conn)
-          pt.setWritable(conn)
-          pt.peerId = pi.id
-          callback(null, pt)
+          proxyConn.setInnerConn(conn)
```
dignifiedquire, line 171: soo much nicer <3

src/identify.js
```diff
@@ -66,23 +66,26 @@ exports.exec = (rawConn, muxer, pInfo, callback) => {
           callback(null, otherPInfo)
         }))
 
-      const obsMultiaddr = rawConn.getObservedAddrs()[0]
+      rawConn.getObservedAddrs((err, addrs) => {
+        if (err) {}
```
dignifiedquire: At least log the error

dignifiedquire: or return?

src/identify.js
```diff
@@ -105,29 +108,32 @@ exports.handler = (pInfo, swarm) => {
           return
         }
         const msg = idPb.Identify.decode(data)
-        if (msg.observedAddr.length > 0) {
+        if (msg.observedAddr && msg.observedAddr.length > 0) {
```
dignifiedquire: this is duplicated throughout, maybe put it into a function 

``` js
function hasObservedAddr (msg) {
  return msg.observedAddr && msg.observedAddr.length > 0 
}
```

src/index.js
```diff
@@ -122,8 +120,16 @@ function Swarm (peerInfo) {
...
+    const transports = this.transports
+
+    parallel(Object.keys(transports).map((key) => {
+      return (cb) => {
+        parallel(transports[key].listeners.map((listener) => {
+          return (cb) => {
+            listener.close(cb)
+          }
+        }), cb)
+      }
```
dignifiedquire, line 132: This can be written a little bit more compact

``` js
parallel(Object.keys(transports).map((key) => (cb) => {
  parallel(
    transports[key].listeners.map((listener) => listener.close.bind(listener)),
    cb
  )
}))
```

src/identify.js
daviddias: This will change with the extraction of `Identify` (happening today or tomorrow in a separate PR), but yes, we can return for now.

src/identify.js
daviddias: 👍

src/transport.js
```diff
@@ -34,59 +41,86 @@ module.exports = function (swarm) {
...
-          callback = noop // this is done to avoid connection drops as connect errors
-          cb(null, conn)
-        }})
-        conn.once('error', () => {
-          callback(new Error('failed to connect to every multiaddr'))
+        const conn = t.dial(multiaddrs.shift())
+
+        conn.once('error', connectError)
+
+        conn.on('connect', () => {
```
dignifiedquire: should be `.once`

src/transport.js
```diff
@@ -34,59 +41,86 @@ module.exports = function (swarm) {
...
+        conn.once('error', connectError)
+
+        function connectError () {
           if (multiaddrs.length === 0) {
             return callback(new Error('failed to connect to every multiaddr'))
           }
           next(multiaddrs.shift())
+        }
+
+        conn.on('connect', () => {
```
dignifiedquire: `.once`

src/transport.js
```diff
@@ -34,59 +41,86 @@ module.exports = function (swarm) {
...
+          conn.removeListener('error', connectError)
```
dignifiedquire, line 81: maybe we can put this behaviour into a function (adding an error listener and a connect listener to catch connect errors on connecting)

src/transport.js
```diff
@@ -98,27 +132,19 @@ module.exports = function (swarm) {
         return callback(new Error(`Trying to close non existing transport: ${key}`))
       }
 
-      transport.close(callback)
+      transport.listeners.forEach((listener) => { listener.close(callback) })
```
dignifiedquire: no need for `{}`

dignifiedquire: also doesn't this mean `callback` is called multiple times??

dignifiedquire: so maybe this:

``` js
parallel(transport.listeners.map((listener) => listener.close.bind(listener)), callback)
```

src/transport.js
```diff
@@ -98,27 +132,19 @@ module.exports = function (swarm) {
...
   return tp.filter(multiaddrs.map((addr) => {
     // webrtc-star needs the /ipfs/QmHash
     if (addr.toString().indexOf('webrtc-star') > 0) {
       return addr
     }
 
-    // ipfs multiaddrs are not dialable so we drop them here
-    if (contains(addr.protoNames(), 'ipfs')) {
-      return addr.decapsulate('ipfs')
-    }
```
dignifiedquire, line 118: why can we drop this?

daviddias, line 118: because now transports accept with ipfs encapsulation and without

src/transport.js
daviddias: great catch, thank you :)

src/dial.js
daviddias, line 172: This is 'pre Identify factor out', after the refactor it should always be `getPeerInfo`. I'm just doing things gradually to not break too many at once.

dignifiedquire, line 172: but you could just change the impl on the connection so it is a setter, then you don't need to break anything:

``` js
function Connection () {
  // ...
}

Object.setProperty(Connection.prototype, 'peerId', {
  set (id) {
    this._peerId = id
    // other things to do
  }
})


```

PR #108, opened 2016-09-13
Async Crypto Endeavour

dignifiedquire:
Ref https://github.com/libp2p/js-libp2p-crypto/pull/10


gulpfile.js
```diff
@@ -20,37 +20,49 @@ let sigS
...
-    swarmA.transport.listen('ws', {}, echo, cb)
+    PeerId.createFromJSON(
+      JSON.parse(
+        fs.readFileSync(
+          path.join(__dirname, './test/test-data/id-1.json'))
+      ),
+      (err, id) => {
+        if (err) {
+          return cb(err)
+        }
```
daviddias, line 31: It would be super beneficial to have a libp2p module `libp2p-peer-gen` that already has 100 or 1000 PeerId generated and so, each time we request 'one peer', it would be super fast because it only had to load it from disk. We would use this everywhere for testing!

test/01-transport-tcp.node.js
```diff
@@ -1,27 +1,48 @@
...
-  before(() => {
-    peerA.multiaddr.add(multiaddr('/ip4/127.0.0.1/tcp/9888'))
-    peerB.multiaddr.add(multiaddr('/ip4/127.0.0.1/tcp/9999'))
-    swarmA = new Swarm(peerA)
-    swarmB = new Swarm(peerB)
+  let peerA
+  let peerB
+
+  before((done) => {
+    utils.createInfos(2, (err, infos) => {
```
daviddias, line 21: It seems you already went ahead and did what I proposed above. Would you like to make it a module?

test/utils.js
```diff
@@ -0,0 +1,30 @@
...
+      PeerInfo.create((err, info) => {
+        if (err) {
+          return cb(err)
+        }
+
+        infos.push(info)
+        cb()
+      })
+    })
+  }
```
daviddias, line 27: Let's make this read from disk instead :)

daviddias (changes requested):

test/utils.js
dignifiedquire: okay, will pull out and make its own module with pre generated 100 ids, and if it needs more generate them on the fly

dignifiedquire: I've left it in here for now, but it has now ids on disk that it tries to read.

gulpfile.js
daviddias: 🛎

daviddias (approved):
Only missing the Sauce Labs testing badge

dignifiedquire:
> Only missing the Sauce Labs testing badge

done

PR #128, opened 2016-11-11
Catch unhandled exception (EADDRINUSE) in listeners

haadcode:
This PR will make sure EADDRINUSE gets bubbled up in the callback chain so that when we try to bind a peer to an address (port) already in use, it will not cause an unhandled exception and blow up.

This fix is needed to be able to test the same behaviour ultimately in js-ipfs.

### DIFF
--- a/src/transport.js
+++ b/src/transport.js
@@ -81,7 +81,7 @@ module.exports = function (swarm) {
         return (cb) => {
           const listener = transport.createListener(handler)
 
-          listener.on('error', cb)
+          listener.once('error', cb)
 
           listener.listen(ma, () => {
             listener.getAddrs((err, addrs) => {
--- a/src/transport.js
+++ b/src/transport.js
@@ -80,6 +80,9 @@ module.exports = function (swarm) {
       const createListeners = multiaddrs.map((ma) => {
         return (cb) => {
           const listener = transport.createListener(handler)
+
+          listener.on('error', cb)
+
           listener.listen(ma, () => {
             listener.getAddrs((err, addrs) => {
               if (err) {
@@ -93,7 +96,11 @@ module.exports = function (swarm) {
         }
       })
 
-      parallel(createListeners, () => {
+      parallel(createListeners, (err) => {
+        if (err) {
+          return callback(err)
+        }
+
         // cause we can listen on port 0 or 0.0.0.0
         swarm._peerInfo.multiaddr.replace(multiaddrs, freshMultiaddrs)
         callback()
--- a/src/transport.js
+++ b/src/transport.js
@@ -84,6 +84,7 @@ module.exports = function (swarm) {
           listener.once('error', cb)
 
           listener.listen(ma, () => {
+            listener.removeListener('error', cb)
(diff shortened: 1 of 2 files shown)


src/transport.js
```diff
@@ -80,6 +80,9 @@ module.exports = function (swarm) {
       const createListeners = multiaddrs.map((ma) => {
         return (cb) => {
           const listener = transport.createListener(handler)
+
+          listener.on('error', cb)
```
dignifiedquire: please 'once' to gurantee cb is only called once

dignifiedquire:
LGTM, @diasdavid please check this as well before we merge

src/transport.js
```diff
@@ -80,6 +80,9 @@ module.exports = function (swarm) {
       const createListeners = multiaddrs.map((ma) => {
         return (cb) => {
           const listener = transport.createListener(handler)
+
+          listener.once('error', cb)
+
```
daviddias, line 85: This event listener needs to unregistered if the event is successful, otherwise a future error (like change of wifi) will make the callback be called a second time with an error

dignifiedquire: ah right, I knew I was missing sth

dignifiedquire:
There is also the question of, should we fail the whole thing just because one listener failed, i.e. I try to listen on three ports, and one fails what should happen?

src/transport.js
haadcode: I don't know how/where to do that

haadcode: And I don't know if we should do that. Shouldn't the consumers know about such errors and be able to react to them as they see appropriate?

The thing is, these errors are unhandled atm which makes it cause an "unhandled exception" in the higher layers, ie. it needs to be caught on a global level (as opposed to block level). The correct way is to make sure there are no unhandled exceptions.

dignifiedquire: > I don't know how/where to do that

You can do this by calling `listener.removeListener('error', cb)`. The position for this is in the callback to `listen` as that means we are happily listening on our port.

daviddias:
@haadcode CR'ed :)

Also, take a look at: https://github.com/ipfs/community/blob/master/js-commit-guidelines.md (important for the changelog)

haadcode:
What is the "scope" of this change for the commit msg? https://github.com/ipfs/community/blob/master/js-commit-guidelines.md#scope

daviddias:
It would a `fix`, such as: `fix: ensure listener errors are propagated for (EADDRINUSE)`

dignifiedquire:
The scope would be `transport` in this case

src/transport.js
daviddias: removing listeners is part of the events API (https://nodejs.org/api/events.html#events_event_removelistener), you essentially just need to passed a named function.

> Shouldn't the consumers know about such errors and be able to react to them as they see appropriate?

There are errors that belong to the 'start listening process' and errors that are 'the socket or the interface where I was listening went away', these need to be handled differently

haadcode:
Fixed

PR #131, opened 2016-11-20
Calling distinctMultiaddr to eliminate appropriate dup multiaddrs

SidHarder:
Calling distinctMultiaddr should solve issue https://github.com/ipfs/js-ipfs/issues/228.  The Daemon shouldn't choke because duplicate ports have been eliminated.


daviddias:
thank you @SidHarder : )

restarted CI with no cache so that it fetches latest peer-info, let's see if we are missing something :)

PR #133, opened 2016-11-29
Added error handling to transport.dial functionality when there is one multiaddr

michaelfakhri:
Hey All,

Dialing an offline node does not currently return an error but simply returns a connection whether or not the connection was established. This PR ensures that it returns an error when there is one multiaddr for a given transport. The reason the same was not done for multiple multiaddrs is that it seems to me that code for multiple multiaddrs is not fully implemented <- will discuss in more detail by referencing code below.

NOTES
- This PR currently fails tests since it requires the signaling error handling code in libp2p/js-libp2p-webrtc-star#39 which is not available in the current release of webrtc-star.
- This PR fixes the race between signaling discovery by changing it into a simple dial error that way it is handled the same way other dial errors are. libp2p/js-libp2p-webrtc-star#5 <- this issue can be closed after this is mergerd

Thanks

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -12,11 +12,16 @@ const multiaddr = require('multiaddr')
 const fs = require('fs')
 const path = require('path')
 
-const sigServer = require('libp2p-webrtc-star/src/signalling-server')
+const sigServer = require('libp2p-webrtc-star/src/signalling')
 
 let swarmA
 let swarmB
 let sigS
+
+const options = {
+  port: 15555,
+  host: '127.0.0.1'
+}
 
 gulp.task('test:browser:before', (done) => {
   function createListenerA (cb) {
@@ -70,7 +75,7 @@ gulp.task('test:browser:before', (done) => {
 
   createListenerA(ready)
   createListenerB(ready)
-  sigS = sigServer.start(15555, ready)
+  sigS = sigServer.start(options, ready)
 
   function echo (protocol, conn) {
     pull(conn, conn)
--- a/package.json
+++ b/package.json
@@ -45,7 +45,7 @@
     "libp2p-secio": "^0.6.3",
     "libp2p-spdy": "^0.10.0",
     "libp2p-tcp": "^0.9.1",
-    "libp2p-webrtc-star": "^0.6.0",
+    "libp2p-webrtc-star": "^0.7.0",
     "libp2p-websockets": "^0.9.1",
     "pre-commit": "^1.1.3",
     "pull-goodbye": "0.0.1",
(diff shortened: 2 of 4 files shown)


michaelfakhri:
Snippet from src/transport.js
```
      // b) if multiaddrs.length = 1, return the conn from the
      // transport, otherwise, create a passthrough
      if (multiaddrs.length === 1) {
        const conn = t.dial(multiaddrs.shift(), (err) => {
          if (err) return callback(err)
          callback(null, new Connection(conn))
        })
        return
      }

      // c) multiaddrs should already be a filtered list
      // specific for the transport we are using
      const proxyConn = new Connection()

      next(multiaddrs.shift())

      // TODO improve in the future to make all the dials in paralell
      function next (multiaddr) {
        const conn = t.dial(multiaddr, () => {
          proxyConn.setInnerConn(conn)
          callback(null, proxyConn)
        })
      }
```
looking at c), the callback is called on the first multiaddr regardless of succes status. If we want generic error handling code, we could simplify b) and c) to the following:
```
      // b) multiaddrs should already be a filtered list
      // specific for the transport we are using

      next(multiaddrs.shift())

      // TODO improve in the future to make all the dials in parallel
      function next (multiaddr) {
        const conn = t.dial(multiaddr, (err) => {
          if (err && multiaddrs.length === 0) return callback(err)
          else if (err) return next(multiaddrs.shift())
          const proxyConn = new Connection()
          proxyConn.setInnerConn(conn)
          callback(null, proxyConn)
        })
      }
```
^ This would provide generic error handling for both cases of b) and c) together which also passes all tests.
Just a suggestion, let me know what you think.

diasdavid:
@michaelfakhri your added test is failing, could you check?

daviddias (changes requested):
@michaelfakhri your added test is failing, could you check?

michaelfakhri:
@diasdavid It needs code from libp2p/js-libp2p-webrtc-star#39 which is not available in latest release v0.6.0 of webrtc-star. It passes locally with these changes manually added. Do you want to bump webrtc-star to v0.6.1 and publish?
Have you had a look at the code in https://github.com/libp2p/js-libp2p-swarm/pull/133#issuecomment-263500622. Is this what the code is supposed to be doing? Or is it something completely different?

daviddias:
@michaelfakhri released :)

michaelfakhri:
Closing and reopening to re-trigger tests

michaelfakhri:
@diasdavid Tests are passing locally on my machine but circleci is failing on a apt-get update command while travis is failing on SAUCE connection. Not sure how do you want to proceed?

daviddias:
@michaelfakhri you need to update to version 0.7.0

michaelfakhri:
@diasdavid done

PR #139, opened 2016-12-14
fix(transport): do not callback after listen errored

dignifiedquire:
Fixes the error seen in https://github.com/haadcode/orbit/issues/210#issuecomment-267102858

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -2,7 +2,8 @@
 
 const util = require('util')
 const EE = require('events').EventEmitter
-const parallel = require('async/parallel')
+const each = require('async/each')
+const series = require('async/series')
 const includes = require('lodash.includes')
 
 const transport = require('./transport')
@@ -84,10 +85,10 @@ function Swarm (peerInfo) {
 
   // Start listening on all available transports
   this.listen = (callback) => {
-    parallel(this.availableTransports(peerInfo).map((ts) => (cb) => {
+    each(this.availableTransports(peerInfo), (ts, cb) => {
       // Listen on the given transport
       this.transport.listen(ts, {}, null, cb)
-    }), callback)
+    }, callback)
   }
 
   this.handle = (protocol, handlerFunc, matchFunc) => {
@@ -124,19 +125,23 @@ function Swarm (peerInfo) {
   }
 
   this.close = (callback) => {
-    Object.keys(this.muxedConns).forEach((key) => {
-      this.muxedConns[key].muxer.end()
-    })
-
-    const transports = this.transports
-
-    parallel(
-      Object.keys(transports).map((key) => (cb) => {
-        parallel(transports[key].listeners.map((listener) => {
-          return (cb) => listener.close(cb)
-        }), cb)
-      }),
-      callback
-    )
+    series([
+      (cb) => each(this.muxedConns, (conn, cb) => {
+        conn.muxer.end((err) => {
+          // If OK things are fine, and someone just shut down
+          if (err && err.message !== 'Fatal error: OK') {
+            return cb(err)
+          }
+          cb()
+        })
+      }, cb),
+      (cb) => {
+        each(this.transports, (transport, cb) => {
+          each(transport.listeners, (listener, cb) => {
+            listener.close(cb)
+          }, cb)
+        }, cb)
+      }
+    ], callback)
   }
 }
--- a/src/transport.js
+++ b/src/transport.js
@@ -2,6 +2,7 @@
 
 const Connection = require('interface-connection').Connection
 const parallel = require('async/parallel')
+const once = reuqire('once')
 const debug = require('debug')
 const log = debug('libp2p:swarm:transport')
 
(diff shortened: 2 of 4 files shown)


src/transport.js
```diff
@@ -82,10 +82,18 @@ module.exports = function (swarm) {
...
+            err = true
+            cb(err)
+          })
 
           listener.listen(ma, () => {
+            // Already errored out
+            if (err) {
+              return
+            }
+
```
daviddias: Good catch!

Can we make it more simpler with

```
listener.once('error', cb)
...
listener.removeListener('error', cb)
```

daviddias (changes requested):

src/transport.js
dignifiedquire: We can't, that's what it was before, which doesn't catch the case where an error happens but the listen callback still happens

michaelfakhri: May I offer an alternative to these two ways?
```
listener.once('error', (err) => {
  if (err !== 'timeout') { // timeout handled internally in libp2p-webrtc-star when connect_error is fired up
    listener.close() // data packet parse error
  }
})
```
Reasoning: socket io client emits two events on connection error (connect_error and error). callback is attached internally to connect error. so no need to attach again for error. BUT error is also called on a data packet parse error, so we need some kind of error handling for the two scenarios. Hence ^

dignifiedquire:
@michaelfakhri thanks but as far as I know this handler at the moment doesn't use socket.io. Only pull-ws, net and webrtc

michaelfakhri:
@dignifiedquire the webrtc transport uses socket.io internally to connect to the signalling server. Isn't that where you are seeing the error, while connecting to the sig server? 
Actually another alternative is to remove the connect_error event handler from the webrtc transport, that way its much cleaner and simpler. since I doubt parse errors are actually the problem here. Are they?

michaelfakhri:
A bit of background: I am seeing the same exact error "callback was already called" while testing the browser application I am building using libp2p. The reason in my case being the restarting of the sig server while changing between mocha -> karma tests. I believe your error would be something similar  relating to the sig server connection.

dignifiedquire:
@michaelfakhri I think I was seeing the issue in other cases as well. But in any case the handling here should not include checking for transport specific errors, if signaling server transport emits errors that shouldn't be handled as such in here, they should not be emitted from there in the first place. 

I've added another important fix for close behaviour. Before we were not waiting for the stream muxer to properly close. Now we do. This depends on https://github.com/libp2p/js-libp2p-spdy/pull/47

michaelfakhri:
@dignifiedquire Yes I completely agree with everything you said.

Just for reference: here is the error I was talking about earlier, it isn't actually caused by the reason I mentioned earlier. I wrote a test that emits the callback was already called error. I think its caused by multiple events firing internally inside socket.io-client. I haven't looked that deep into it but I can tell something quite weird is going on and I dont think its any of the code in here that's directly causing this error.

https://github.com/michaelfakhri/js-libp2p-swarm/commit/0d1f8bf7bc0a9bf397ed0add21643201da40ad9d

src/index.js
```diff
@@ -124,19 +124,18 @@ function Swarm (peerInfo) {
...
+    }, (err) => {
+      if (err) {
+        return callback(err)
+      }
+      const transports = this.transports
+      each(Object.keys(transports), (key, cb) => {
+        each(transports[key].listeners, (listener, cb) => {
+          listener.close(cb)
+        }, cb)
+      }, callback)
```
daviddias: Why can't all they be closed in parallel?

daviddias (approved):
Small question, otherwise LGTM

src/index.js
dignifiedquire: `each` is parallel

daviddias: Interesting, thank you :)

daviddias (approved):
Let's ensure CI is happy before merging

dignifiedquire:
Yeah looks like one test is unhappy, fixing now.

dignifiedquire:
should be happy now

daviddias:
@dignifiedquire CI still says `no`

dignifiedquire:
@diasdavid I don't get it :/ all passing locally

daviddias:
@dignifiedquire the only different to CI, is that CI is doing a fresh npm install every time, have you tried that in your local machine?

dignifiedquire:
Yeah I did, could you try as well?

daviddias:
Locally, it works for me too.

So, the error is:

```
  1) libp2p-swarm high level API - with everything mixed all together! close a muxer emits event:
     Uncaught AssertionError: expected true to not exist
      at conn.getPeerInfo (test/09-swarm-with-muxing.node.js:214:27)
```

But that line is on the `dial from tcp+ws to tcp+ws`, meaning that conn.getPeerInfo is being resolved after the hangup (CI is slower on things).

dignifiedquire:
@diasdavid ready for squash and ship

daviddias:
squashed and shipped :)

PR #158, opened 2017-01-23
wip

daviddias:


src/connection.js
```diff
@@ -31,15 +31,31 @@ module.exports = function connection (swarm) {
...
+            console.log('waterfall')
             waterfall([
-              (cb) => ms.handle(conn, cb),
-              (cb) => ms.select(identify.multicodec, cb),
-              (conn, cb) => identify.dialer(conn, cb),
+              (cb) => {
+                console.log('ms-handle')
+                console.log(conn)
+                console.log('\n')
+                setTimeout(() => ms.handle(conn, cb), 2000)
```
daviddias: @VictorBjelkholm found out that spdy also fails, if we have a set timeout, there is a race condition on this, it was working for spdy by chance.

daviddias: nvm, I just didn't increase the timeout on spdy for the checking. Still remains an issue with multiplex.

PR #160, opened 2017-01-24
Don't skip multiplex identify test anymore

victorb:
If https://github.com/libp2p/js-libp2p-multiplex/pull/55 gets merged, this test passes again.

Depends on the PR referenced above


daviddias:
same as https://github.com/libp2p/js-libp2p-swarm/pull/159

PR #163, opened 2017-01-26
Use callback in muxer.end instead of close event

victorb:
Solves https://github.com/ipfs/js-libp2p-ipfs-nodejs/issues/58

Not sure if this is the right way to fix it though, since multiplex seems to not correctly emit close event...

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -114,11 +114,11 @@ function Swarm (peerInfo) {
     const key = peerInfo.id.toB58String()
     if (this.muxedConns[key]) {
       const muxer = this.muxedConns[key].muxer
-      muxer.end()
       muxer.once('close', () => {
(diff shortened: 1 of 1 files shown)


src/index.js
```diff
@@ -114,8 +114,7 @@ function Swarm (peerInfo) {
     const key = peerInfo.id.toB58String()
     if (this.muxedConns[key]) {
       const muxer = this.muxedConns[key].muxer
-      muxer.end()
-      muxer.once('close', () => {
```
daviddias, line 118: Try changing the order of these too. This event should be emitted

victorb:
@diasdavid you were right, changing order fixes issue as well!

daviddias (approved):

PR #193, opened 2017-03-26
Attempt to dial all addresses in parallel and use the first success

jackkleeman:
https://github.com/ipfs/js-ipfs/issues/812
We were not dialling all addresses in sequence, currently only the first address was ever dialled leading to regular failures. This dials all in parallel and uses the first success (or the first timeout).

### DIFF
--- a/src/transport.js
+++ b/src/transport.js
@@ -2,6 +2,7 @@
 
 const Connection = require('interface-connection').Connection
 const parallel = require('async/parallel')
+const each = require('async/each')
 const once = require('once')
 const debug = require('debug')
 const log = debug('libp2p:swarm:transport')
@@ -53,11 +54,18 @@ module.exports = function (swarm) {
       // specific for the transport we are using
       const proxyConn = new Connection()
 
-      next(multiaddrs.shift())
+      each(multiaddrs, next)
 
-      // TODO improve in the future to make all the dials in paralell
+      var success = false
       function next (multiaddr) {
+        if (success) {
+          return;
+        }
         const conn = t.dial(multiaddr, () => {
+          if (success) {
+            return;
+          }
+          success = true;
           proxyConn.setInnerConn(conn)
           callback(null, proxyConn)
         })


src/transport.js
```diff
@@ -53,11 +54,18 @@ module.exports = function (swarm) {
       // specific for the transport we are using
       const proxyConn = new Connection()
 
-      next(multiaddrs.shift())
+      each(multiaddrs, next)
```
dignifiedquire, line 57: I think you could use [`async.race`](http://caolan.github.io/async/docs.html#race) instead, then you don't have to do the manual success tracking

dryajov: Won't `async.race` fail on the first failed attempt? I think we'd want all the addresses to be dialed and be able to inspect those that worked, and not stop on first fail - a la bluebird inspect way, `async.parallel` seems to do just that? Also, not sure if dialing in parallel is the best course of action either, unless we want to open connections on all available transports? At the very least, it should try ALL the ~~transports~~ addresses until one works, the fact that it didn't is a bug.

jackkleeman: Currently i think it just takes the first return, which is generally either a success or a timeout, which works well for me. I think dialling in parallel is needed because websocket timeout is around ten seconds, so we really want to try every address at once and then move forward with whatever sticks.

dryajov: Mmm... that makes sense. 

Does it makes sense to keep all the available connections around? Would it makes sense to switch between connection depending on latency (probably overkill)?

In fairness tho, the first one to come back is probably the fastest one already 🍭

jackkleeman: My thinking exactly

dryajov: @dignifiedquire @jackkleeman don't think that either `race` or `parallel` would do what we want in this case. With `race` if the first result is a failure, the callback is invoked even if others might succeed, with `parallel` the callback would be called only after all of the calls are completed, I think we need [async.some](https://caolan.github.io/async/docs.html#some) here 🎢

dignifiedquire: I think the proper way to do it is probably this:

- use async.queue with a concurrency of e.g 4
- push all addrs as task to the queue
- start the queue
- in the worker function check if success and if so call `queue.kill` and the callback
- if after finishing the queue no success happened call back in with an error

dryajov: @dignifiedquire that should work. I like that we can control the concurrency and not have 20 (figurative and literal 😄 ) dials happen at once.

jackkleeman: Cool, I'll leave it up to you to make this change as I'm not well versed in the async functions. I'm glad I could bring the issue to the table :)

dryajov: @jackkleeman @dignifiedquire I'm deep into libp2p for the circuit-relay stuff right not, so I can squeeze this in, how quickly do you need this @jackkleeman? Is this something thats blocking you?

jackkleeman: It's not blocking me at all as I already maintain a fork of this package for my project (OpenBazaar in the browser), so i will use my simple async.each solution in my fork, and leave it up to you to make a better fix whenever you have time. I would argue it's worth using async.each here in the meantime, as the current implementation is fairly unusable unless you manually filter addresses to be valid, but that's totally up to you guys.

jackkleeman: Oh and i don't want you to slow down on the relay as it will be game changing for my project!!

dryajov: @jackkleeman - yep, that was a great catch overall, I've patched my local fork with the simpler sequential `next(shift)` so far. This might have been the culprit of a bunch of weird issues I was running into at some point with the the WS vs TCP order of dialing https://github.com/libp2p/js-libp2p-circuit/pull/1#issuecomment-286959112 🥇

dryajov: @diasdavid @dignifiedquire +1 to merging this as is for now. 

I've created an issue to track the enhancement proposed by @dignifiedquire

src/transport.js
```diff
@@ -53,11 +54,18 @@ module.exports = function (swarm) {
...
 
-      // TODO improve in the future to make all the dials in paralell
+      var success = false
       function next (multiaddr) {
+        if (success) {
+          return;
+        }
         const conn = t.dial(multiaddr, () => {
+          if (success) {
+            return;
```
daviddias, line 66: This is leaving open sockets behind. We need to check: `If I have successfully opened a socket before, then I need to close this one.`

daviddias (changes requested):
Thank you for detecting this and PR'ing a fix @jackkleeman. Just a small comment that needs to be addressed, otherwise LGTM.

src/transport.js
daviddias: See: https://github.com/libp2p/js-libp2p-swarm/pull/193#discussion_r108138054

dignifiedquire: I am going to implement the new dialing today as I need to go into that code anyway

src/transport.js
jackkleeman: If @dignifiedquire is replacing this PR today should I ignore this?

dignifiedquire: Feel free to ignore this yes, I will ping you once the new implementation is up

jackkleeman: Thanks! As it turns out, my each solution is failing in a few cases already (when one address fails immeadiately instead of slowly timing out)

jackkleeman:
Outdone by #195

PR #195, opened 2017-03-27
feat(transport): use parallel limited dialer

### LINKED ISSUE #194: rework transport.dial to dial addresses in parallel
- use async.queue with a concurrency of e.g 4
- push all addrs as task to the queue
- start the queue
- in the worker function check if success and if so call queue.kill and the callback
- if after finishing the queue no success happened call back in with an error

Initially discussed here - https://github.com/libp2p/js-libp2p-swarm/pull/193
Needed to fix this - https://github.com/ipfs/js-ipfs/issues/812

dignifiedquire:
cc @jackkleeman @dryajov @diasdavid 

I haven't added specific test cases, if you have some please drop them in the comments so I can add them.

Needs
- https://github.com/libp2p/js-libp2p-tcp/pull/77
- https://github.com/libp2p/js-libp2p-websockets/pull/60

Replaces #193

Fixes #194


src/transport.js
```diff
@@ -36,32 +42,76 @@ module.exports = function (swarm) {
...
+        log('setting up new queue')
+        q = queue((multiaddr, cb) => {
+          const conn = t.dial(multiaddr, (err) => {
+            if (err) {
+              log('dial failed: %s', multiaddr.toString())
+              return cb(err)
+            }
+            if (q.canceled) {
+              log('dial canceled: %s', multiaddr.toString())
+              // clean up already done dials
```
daviddias: Where is the conn being closed, if we already had dialed one?

dignifiedquire: hahahaha I forgot to write it

dignifiedquire: fixed

src/transport.js
```diff
@@ -36,32 +43,76 @@ module.exports = function (swarm) {
...
-      }
+      // create dial queue if non exists
+      let q
+      if (queues.has(key)) {
+        log('reusing queue')
+        q = queues.get(key)
+      } else {
+        log('setting up new queue')
+        q = queue((multiaddr, cb) => {
+          const conn = t.dial(multiaddr, (err) => {
```
jackkleeman: does dial return an error to its callback? It doesn't seem to for the websocket transport

src/transport.js
dignifiedquire: probably should use `.close` instead

src/transport.js
```diff
@@ -36,32 +43,76 @@ module.exports = function (swarm) {
...
+            if (err) {
+              log('dial failed: %s', multiaddr.toString())
+              return cb(err)
+            }
+            if (q.canceled) {
+              log('dial canceled: %s', multiaddr.toString())
+              // clean up already done dials
+              pull(pull.empty(), conn)
```
daviddias: The other end will try to write to this conn as soon as it opens and since we are not reading those values, it will stay forever open.

Use .close/destroy and make sure it gets destroyed (we don't want to exhaust open sockets/fd limits)

src/transport.js
daviddias: yeah, https://github.com/libp2p/js-libp2p-swarm/pull/195/files#r108174470

src/transport.js
dignifiedquire: fixing this

src/transport.js
dignifiedquire: yeah, working on it

src/transport.js
jackkleeman: I think I am finding a bug where the onConnect callback is being called even when the websocket 'open' event isn't called. Will file an issue. EDIT - actually scratch that, the problem is the websocket transport isn't handling errors correctly.

jackkleeman: https://github.com/pull-stream/pull-ws/issues/18 it was upstream of libp2p.

jackkleeman: See https://github.com/libp2p/js-libp2p-websockets/pull/60
After this PR, websockets will indeed return an error to its callback, so your code will work great 👍

dignifiedquire: Okay I got everything working, except that we need to upgrade both libp2p-tcp and libpp2-websocket to actually provide an error if one happened on the callback, otherwise the detection is horrific

dignifiedquire: And the websocket side depends on the pull-ws issue

dignifiedquire:
I've also added a timeout on the actual dial call so we don't hang indefinitely

src/transport.js
jackkleeman: No there wasn't a pull-ws issue after all, just a libp2p-websockets error handling issue
and I've issued a PR to libp2p-websockets, see my comment above

dignifiedquire: well it is both I am afraid, I actually can't detect the error in libp2p-websockets easily as the connect callback is called even though an error occured :(

jackkleeman (approved):

src/transport.js
jackkleeman: I think onConnect is meant to be called with errors as the first argument. Why is that difficult to check for?

jackkleeman: I opened then closed an issue on this as I realised that they sent errors to onConnect in several places. https://github.com/pull-stream/pull-ws/issues/18

dignifiedquire: Just merged you pr, thank you it works great with that :)

daviddias:
Seems that tests in swarm didn't catch all the cases, specially when there is WS + TCP involved. Getting these at libp2p-ipfs-nodejs:

```
 1) libp2p-ipfs-nodejs libp2p.dial using PeerInfo nodeA to nodeB:
     Uncaught TypeError: conn.close is not a function
      at dialWithTimeout (node_modules/libp2p-swarm/src/transport.js:70:22)
      at injectedCallback (node_modules/async/timeout.js:62:30)
      at transport.dial (node_modules/libp2p-swarm/src/transport.js:199:7)
      at f (node_modules/once/once.js:25:25)
      at Socket.rawSocket.once (node_modules/libp2p-tcp/src/index.js:40:7)
      at TCPConnectWrap.afterConnect [as oncomplete] (net.js:1074:10)
```

TCP is missing the conn.close https://github.com/libp2p/js-libp2p-tcp/blob/master/src/index.js and also, WebRTC star doesn't have such a method https://github.com/libp2p/js-libp2p-webrtc-star/blob/master/src/index.js

Seems that a `.close` never really made it to the interface-connection (https://github.com/libp2p/interface-connection/blob/master/src/connection.js) and with good reason, it has a method to 'end' the call by sending the 'FIN' packet which gets fired once we do something like `pull.empty()`, needing only a 'destroy' event that would trickly to the socket.

daviddias:
Will finish the update on the new libp2p API for bitswap and js-ipfs and come back to this after that :)

dignifiedquire:
@diasdavid so should I simply change it to what I had before? `pull(pull.empty(), conn)`?

daviddias:
Not quite the ideal solution, it has worked for us because we always end muxed streams and only close sockets when they get disconnected. Now we actually need to 'destroy' the socket.

PR #201, opened 2017-03-27
feat: adding relay (#181)

daviddias:


dryajov:
Couldn't push to this branch, so here is a PR with the latest changes - #204

dryajov:
@diasdavid @dignifiedquire 

This PR: #216 is already merged here because I need this functionality in circuit, but it is its own PR.

daviddias:
@dryajov mind squashing the commits on this branch? The commit history is a bit messy making it harder to review. We have the best practice of using `rebase` and not `merge`.

dryajov:
@diasdavid will do!

daviddias:
@dryajov could you take this PR, #216 and #217 and coalesce it all in one single PR submitted by you, it will make the review way simpler for both of us :)

dryajov:
@diasdavid both #216 and #217 are already merged with this PR, we should be able to safely close those.

src/connection.js
```diff
@@ -55,7 +57,7 @@ module.exports = function connection (swarm) {
             }
             const b58Str = peerInfo.id.toB58String()
 
-            swarm.muxedConns[b58Str] = { muxer: muxedConn }
+            swarm.muxedConns[b58Str] = {muxer: muxedConn}
```
daviddias: unnecessary cosmetic change

src/connection.js
```diff
@@ -92,6 +94,11 @@ module.exports = function connection (swarm) {
       })
     },
 
+    relay (config) {
+      swarm.relay = true
+      swarm.transport.add(Circuit.tag, new Circuit.Dialer(swarm, config))
```
daviddias: I rather prefer to make relay specific and not a 'transport' like this. The transport interface will evolve and that doesn't necessarily need to trickle to relay. 

On the dial.js, after trying all transports, then it should try relay to get that first conn.

src/index.js
```diff
@@ -6,11 +6,14 @@ const each = require('async/each')
 const series = require('async/series')
 const transport = require('./transport')
 const connection = require('./connection')
+const handler = require('./handler')
 const dial = require('./dial')
 const protocolMuxer = require('./protocol-muxer')
 const plaintext = require('./plaintext')
 const assert = require('assert')
 
+const DEFAULT_TRANSPORT_PRIORITY = 1
```
daviddias: This should be removed now.

daviddias:
As I mentioned here:
- https://github.com/libp2p/js-libp2p-circuit/pull/9/files#r116323578
- https://github.com/libp2p/js-libp2p-circuit/pull/9/files#r115850926

The relay listener never gets used.

dryajov:
~~@diasdavid since we're not doing multihop dialing anymore and this was needed specifically to support it, should we close this PR?~~

@diasdavid We still need this for listening on specific relays, overall I think this makes sense as it allows extending and plugin with the swarm more easily.

dryajov:
@diasdavid @lgierth @dignifiedquire Closing this since we're not doing multihop dialing anymore.

PR #203, opened 2017-03-27
Fix/dial class

dignifiedquire:


src/dialer.js
```diff
@@ -0,0 +1,185 @@
...
+   * @param {CancelToken} token
+   * @param {function(Error, Connection)} callback
+   * @returns {void}
+   */
+  push (transport, addr, token, callback) {
+    this.queue.push({transport, addr, token}, callback)
+  }
+}
+
+module.exports = Dialer
```
daviddias: Let's avoid putting several classes in the same file, we already learned how that ends from Java.

src/dial.js
```diff
@@ -87,8 +87,7 @@ module.exports = function dial (swarm) {
       nextTransport(tKeys.shift())
 
       function nextTransport (key) {
-        const multiaddrs = pi.multiaddrs.slice()
-        swarm.transport.dial(key, multiaddrs, (err, conn) => {
+        swarm.transport.dial(key, pi, (err, conn) => {
```
daviddias, line 90: Hm, I intentionally had made `swarm.transport.dial` be just multiaddr aware for simplicity, and so, it would only receive a multiaddr and not a peerInfo

swarm.transport.dial -> deals with dialing on a transport
swarm.dial -> does all the magic of upgrading the connection

daviddias: This is kind of a breaking change on the API, not a biggie because it is more of an internal that is just exposed for testing, however, I'm not sure if I see the advantage, could you clarify?

daviddias (changes requested):

src/dialer.js
```diff
@@ -0,0 +1,185 @@
...
+      (err, conn) => {
+        if (err) {
+          log('dialQueue:work:error')
+          return callback(null, {error: err})
+        }
+
+        if (token.cancel) {
+          log('dialQueue:work:cancel')
+          // clean up already done dials
+          pull(pull.empty(), conn)
```
daviddias: Just to make sure we don't loose track of this:

- https://github.com/libp2p/js-libp2p-swarm/pull/200#discussion_r108214454
- https://github.com/libp2p/js-libp2p-swarm/pull/200#issuecomment-289577321

daviddias: https://github.com/libp2p/js-libp2p-swarm/issues/205

src/dialer.js
dignifiedquire: lol

src/dial.js
dignifiedquire: I am doing the same thing that go does, that is limiting dials per peer. To do that I need to know the peerId otherwise I can't track dials per peer

daviddias: ok, we can keep it.

src/dialer.js
daviddias: ok, just missing this then :)

dignifiedquire: done

daviddias (approved):

PR #204, opened 2017-03-28
Feat/adding relay

dryajov:
Can't push to the branch libp2p:feat/adding-relay, so here is a PR with the latest changes

PR #216, opened 2017-04-02
Feat/expose connection handling flow in dial

dryajov:
@diasdavid @dignifiedquire 

I need the ability to pass a raw connection to the swarm and have it encrypted and muxed, this is required for onion dialing, since when I get a conn from one relay to another relay, I have no way to upgrade that with encryption and muxing (handshake).

Here is a typical flow:

* having a chained multiaddr like this `/ipfs/QmRealy1/p2p-circuit/ipfs/QmRelay2/p2p-circuit/ipfs/QmDest`
   * I need to dial into `/ipfs/QmRealy1/` and hand it the `/ipfs/QmRelay2` address as a dest
      * at this point, a `relay1 <=> relay2` circuit exists, but since it hasn't been returned to the swarm it has not gone through muxing and encryption. Next time I read from that connection, I get a multristream header sent from QmRelay2 that is the start of the regular `p2p` handshake, for the `relay1 <=> relay2` connection.

Exposing the the dial/handshake flow would allow me to manually send the  `relay1 <=> relay2` to the swarm and have it go through the handshake. The other option would be to reimplement the handshake in circuit, but I don't think we want to do that :)


daviddias:
Is this still being used? @dryajov If yes, could you coalesce this PR with #204 for me to review?

Thank you!

dryajov:
This is already part of #181 and can be safely closed.

PR #217, opened 2017-04-03
feat: sort available transports by priority

dryajov:
This uses the the priority property added in the bellow PRs to prioritize order of transport dialing.

https://github.com/libp2p/js-libp2p-tcp/pull/78 (see this for discussion)
https://github.com/libp2p/js-libp2p-websockets/pull/61

If we decide to go with this, we'd need to add `priority` to the remaining transports as well.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,10 +51,11 @@
     "libp2p-tcp": "~0.10.0",
     "libp2p-webrtc-star": "~0.9.0",
     "libp2p-websockets": "~0.10.0",
+    "peer-book": "~0.4.0",
     "pre-commit": "^1.2.2",
     "pull-goodbye": "0.0.1",
-    "peer-book": "~0.4.0",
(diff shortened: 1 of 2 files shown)


dryajov:
After this commit (04237446784ea8e06ca7e0307a24243d601d5add), we no longer need the related changes in each transport.

daviddias:
Hi @dryajov. This sounds like a good idea in the long term but priorities essentially mean nothing until we expose these features to the user to configure. Setting default priorities is exactly the same as prioritizing direct connections and putting on second plan the relayed ones, which makes sense to do.

Let's avoid yet adding one more piece to the interface-transport abstraction until we actually need it. The sort function is: direct connections first, relayed second.

dryajov:
@diasdavid I'll remove the `priority` in favor of pushing circuit addresses to be dialed last.

PR #218, opened 2017-04-06
fix: move emitters to last thing in the method

daviddias:


src/connection.js
```diff
@@ -69,13 +69,14 @@ module.exports = function connection (swarm) {
...
 
-            swarm.emit('peer-mux-established', peerInfo)
             muxedConn.on('close', () => {
               delete swarm.muxedConns[b58Str]
               peerInfo.disconnect()
               peerInfo = swarm._peerBook.put(peerInfo)
               swarm.emit('peer-mux-closed', peerInfo)
             })
+
+            swarm.emit('peer-mux-established', peerInfo)
```
dignifiedquire: lets use `setImmediate` as discussed

daviddias: 👍

PR #220, opened 2017-07-05
fix: move 'pull-stream' from devDependencies to dependencies

ArnaudValensi:
'pull-stream' package is needed in dependencies because it is used in './src/limit-dialer/queue.js'.


daviddias (approved):

daviddias:
thanks @ArnaudValensi ! :)

PR #223, opened 2017-08-16
Feat/adding relay

dryajov:


daviddias:
@dryajov I'm confused if you actually wanted to close this PR or not

dryajov:
@diasdavid this is the one that modified the dialer and moves it into a handler, this was done to support multihop dialing. I'm going to open a new PR shortly with only the circuit related logic present.

PR #224, opened 2017-08-17
feat: Circuit Relay

dryajov:


src/limit-dialer/index.js
```diff
@@ -37,6 +38,7 @@ class LimitDialer {
     log('dialMany:start')
     // we use a token to track if we want to cancel following dials
     const token = { cancel: false }
+    callback = once(callback) // only call callback once
```
dryajov, line 41: @dignifiedquire since `dialMany` will possibly succeed for more than one address, the callback could resolve multiple times, this breaks in a couple of places (I'll have to see exactly where, but I believe it was either libp2p or swarm). I think having the callback called only once, for the first address that gets resolved, makes sense here?

dignifiedquire: If we expect that, it should be set up such that the first one results in the callback being called, but in addition made sure that all others are canceled and no new ones are started to avoid waste

dryajov: Hmm, good idea. Would you mind if I leave this PR as is and open an issue to enhance `dialMany` in another issue? I don't want to hold this PR because of it.

dryajov:
This is currently blocked by - https://github.com/webpack/node-libs-browser/pull/63

daviddias (changes requested):
Needs docs on how to enable the Relay

daviddias:
@dryajov mind updating this PR with all the released deps and make CI happy?

dryajov:
Will do!

dryajov:
@diasdavid I've fixed tests and updated the docs.

daviddias (approved):

PR #225, opened 2017-08-23
Fix callback bug

mkg20001:
The top-level callback was called instead of the supplied cb callback


daviddias (approved):
Thanks @mkg20001 :D

PR #227, opened 2017-08-28
Add test for multiple port 0

victorb:
Current state:

screen shot 2017-08-28 at 14 02 12

### DIFF
--- a/test/transport-tcp.js
+++ b/test/transport-tcp.js
@@ -147,6 +147,25 @@ describe('transport - tcp', () => {
     }
   })
 
+  it('supports multiple addresses with port 0', (done) => {
+    const ma = '/ip4/127.0.0.1/tcp/0'
+    let swarm
+    peer.multiaddrs.add(ma)
+    peer.multiaddrs.add(ma)
+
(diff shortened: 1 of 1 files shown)


victorb:
Found something that is stopping two identical addresses from being added: https://github.com/libp2p/js-peer-info/blob/422d0fd37ec5e34d6af00eb0d03630b8add37a72/src/multiaddr-set.js#L16-L18

This check is probably too early as two addresses with port 0 should be allowed

daviddias:
@VictorBjelkholm interesting. That is definitely a bummer. Probably there should be a clause that for the same network interface, there can be multiple `port 0`. Wanna push a patch?

victorb:
@diasdavid yeah, tried just commenting out that part linked in https://github.com/libp2p/js-libp2p-swarm/pull/227#issuecomment-325389200 but was not the only place. Still working on getting something functional together.

daviddias:
@VictorBjelkholm are you still working on this? Should we open it for contribution?

victorb:
@diasdavid yes, we need this for isolating tests so we'll need to soon-ish. Open up for contribution :+1:

PR #233, opened 2017-10-14
Changed let to var.

alexsicart:
I changed let to var, because when using React/Angular for production, gives a lot of problems with ES6 code.


daviddias:
We use `let` a lot. Why is there a specific problem with this one?

PR #234, opened 2017-10-26
fix: don't dial on relay if not enabled

dryajov:
swarm should not try to dial over circuit if it hasn't been enabled


dryajov:
@diasdavid can we get this one merged as well, some tests are failing due to it in js-ipfs.

daviddias (approved):

PR #236, opened 2017-11-18
fix: don't dial circuit if no transports available

dryajov:
Currently, if there are no transports registered `libp2p-swarm` will still try to dial over circuit relay. This PR adds an error to signal that no protocols are registered for dialing.


dryajov:
@diasdavid perhaps not enabling circuit at all when there are no transports registered is also required, but IMHO this is the safest way of catching all the edge cases.

daviddias (approved):

PR #240, opened 2017-12-07
[WIP] Feat/circuit

dryajov:
Fixing circuit dialing

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,15 +58,15 @@
     "async": "^2.6.0",
     "debug": "^3.1.0",
     "interface-connection": "~0.3.2",
-    "ip-address": "^5.8.8",
+    "ip-address": "^5.8.9",
     "libp2p-circuit": "~0.1.4",
     "libp2p-identify": "~0.6.1",
     "lodash.includes": "^4.3.0",
     "multiaddr": "^3.0.1",
     "multistream-select": "~0.14.1",
     "once": "^1.4.0",
-    "peer-id": "~0.10.2",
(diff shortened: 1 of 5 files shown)


dryajov:
@VictorBjelkholm circle is failing with `sudo dpkg -i libnss3*.deb returned exit code 1`, any ideas what might be causing this?

package.json
```diff
@@ -51,22 +51,23 @@
     "peer-book": "~0.5.1",
     "pre-commit": "^1.2.2",
     "pull-goodbye": "0.0.2",
+    "pump": "^2.0.0",
```
daviddias: Why add an unused dep?

daviddias (changes requested):
What was the fix? Can you add tests to make sure it doesn't happen again?

dryajov:
tests are on the way.

package.json
dryajov: Hmm, it was failing to run on ci because apparently https://github.com/libp2p/js-libp2p-multiplex/blob/fa78df420ca5762b2a003e658fd74bdf1dd3e6b3/src/index.js#L9 is failing to include it as a dep - https://github.com/libp2p/js-libp2p-multiplex/blob/master/package.json#L31..L48. I'll patch `js-libp2p-multiplex`.

dryajov: this should fix the pump issues, as well as port multiplex to use the new aegir https://github.com/libp2p/js-libp2p-multiplex/pull/70

daviddias (approved):

PR #243, opened 2018-02-16
observe traffic and expose statistics

pgte:
Required because of https://github.com/libp2p/js-libp2p/issues/157


test/stats.node.js
```diff
@@ -82,19 +82,47 @@ describe('Stats', () => {
     })
   })
 
-  it('waits a bit', (done) => setTimeout(done, 1900))
+  it('waits a bit', (done) => setTimeout(done, 1000))
```
victorb: A test case that just waits? Should fix the tests to not depend on the ordering...

pgte: @VictorBjelkholm good call! Fixed by moving the preparation code to a before hook.

victorb: @pgte better to get rid of the after/before hooks we have and have function calls in the test cases that returns what you need. If all tests where shaped like this, we can run tests in parallel and just run one isolated test case if wanted.

pgte: In my case I need something to run once, not for every test.
In this case, I want to setup some switches, exchange some protocol-specific data, wait for the stats to compute, and *then* I can infer all the statistics are there.
@VictorBjelkholm Any recommendations?

pgte: @VictorBjelkholm changed to isolated tests for stats, would love your feedback.

test/stats.node.js
```diff
@@ -18,131 +19,188 @@ const tryEcho = utils.tryEcho
...
-
-    switchA.connection.crypto(secio.tag, secio.encrypt)
-    switchB.connection.crypto(secio.tag, secio.encrypt)
 
-    switchA.connection.addStreamMuxer(multiplex)
-    switchB.connection.addStreamMuxer(multiplex)
+      const peerA = infos[0]
+      const peerB = infos[1]
+
+      peerA.multiaddrs.add('/ip4/127.0.0.1/tcp/9001')
```
victorb: This is gonna conflict if we run two test cases at the same time, is it possible to set the port to `0` and then later get it (if you need it) instead? Think you're not actually using the port in the test, so should work fine to be set to `0.`

pgte: @VictorBjelkholm good call, fixed!

victorb (approved):
Tests LGTM :+1:

hacdias (approved):
Looks Awesome 😄

README.md
```diff
@@ -30,6 +30,7 @@ libp2p-switch is used by [libp2p](https://github.com/libp2p/js-libp2p) but it ca
   - [`switch.stop(callback)`](#swarmclosecallback)
   - [`switch.connection`](#connection)
   - [Internal Transports API](#transports)
+  - [`switch.stats`](#stats-api)
```
daviddias: move this above the Internal Transports line.

README.md
```diff
@@ -51,7 +52,7 @@ libp2p-switch is used by [libp2p](https://github.com/libp2p/js-libp2p) but it ca
 ```JavaScript
 const switch = require('libp2p-switch')
 
-const sw = new switch(peerInfo [, peerBook])
+const sw = new switch(peerInfo [, peerBook [, options]])
```
daviddias: This makes it look like options only exists in the context of passing peerBook. What about changing this to just a single options object and assert on its properties?

README.md
```diff
@@ -223,6 +224,106 @@ To avoid the confusion between connection, stream, transport, and other names th
 - connection - something that implements the transversal expectations of a stream between two peers, including the benefits of using a stream plus having a way to do half duplex, full duplex
 - transport - something that as a dial/listen interface and return objs that implement a connection interface
 
+### Stats API
```
daviddias: Do you see this all RED too? 
image

Mind checking what went wrong here?

README.md
```diff
@@ -223,6 +224,106 @@ To avoid the confusion between connection, stream, transport, and other names th
...
+
+#### Global stats
+
+##### `switch.stats.global.snapshot`
+
+Should return a stats snapshot, which is an object containing the following keys and respective values:
+
+* dataSent: amount of bytes sent, [Big](https://github.com/MikeMcl/big.js#readme) number
+* dataReceived: amount of bytes received, [Big](https://github.com/MikeMcl/big.js#readme) number
```
daviddias: `s/*/-/g` for consistency

README.md
```diff
@@ -223,6 +224,106 @@ To avoid the confusion between connection, stream, transport, and other names th
...
+
+Each one of these values is [an exponential moving-average instance](https://github.com/pgte/moving-average#readme).
+
+#### Per-transport stats
+
+##### `switch.stats.transports()`
+
+Returns an array containing the tags (string) for each observed transport.
+
+##### `switch.stats.forTransports(transportTag).snapshot`
```
daviddias: nickpick `s/forTransports/transport/g` (singular and no need the word for, I think)

package.json
```diff
@@ -56,12 +56,14 @@
...
   "dependencies": {
     "async": "^2.6.0",
+    "big.js": "^5.0.3",
     "debug": "^3.1.0",
     "interface-connection": "~0.3.2",
     "ip-address": "^5.8.9",
     "libp2p-circuit": "~0.1.4",
     "libp2p-identify": "~0.6.3",
     "lodash.includes": "^4.3.0",
+    "moving-average": "^1.0.0",
```
daviddias, line 66: I'm glad you created this module :)

src/connection.js
```diff
@@ -25,16 +24,15 @@ module.exports = function connection (swarm) {
       swarm.handle(muxer.multicodec, (protocol, conn) => {
         const muxedConn = muxer.listener(conn)
 
-        muxedConn.on('stream', (conn) => {
-          protocolMuxer(swarm.protocols, conn)
-        })
+        muxedConn.on('stream', swarm.protocolMuxer(protocol))
```
daviddias: my bad for not renaming things here. Mind s/swarm/switch/g all the occurrences?

src/connection.js
```diff
@@ -25,16 +24,15 @@ module.exports = function connection (swarm) {
...
 
         // If identify is enabled
         //   1. overload getPeerInfo
         //   2. call getPeerInfo
         //   3. add this conn to the pool
         if (swarm.identify) {
           // overload peerInfo to use Identify instead
+          const setPeerInfo = conn.setPeerInfo
```
daviddias: Isn't there an overload of setPeerInfo? I'm afraid that this might cause issues in the future. Is it needed?

src/connection.js
```diff
@@ -120,7 +123,7 @@ module.exports = function connection (swarm) {
       swarm.handle(tag, (protocol, conn) => {
         const myId = swarm._peerInfo.id
         const secure = encrypt(myId, conn, undefined, () => {
-          protocolMuxer(swarm.protocols, secure)
+          swarm.protocolMuxer(tag)(secure)
```
daviddias: Something is not correct here. Doing swarm.protocolMuxer(tag) is saying "After you get the conn encrypted, you can encrypt again". It should be for all the Protocols being handled and not just for one.

Note: I'm still reading through the PR and you might have changed something else.

src/protocol-muxer.js
```diff
@@ -1,21 +1,29 @@
...
+    const ms = new multistream.Listener()
 
-  Object.keys(protocols).forEach((protocol) => {
-    if (!protocol) {
-      return
-    }
+    Object.keys(protocols).forEach((protocol) => {
+      if (!protocol) {
+        return
+      }
```
daviddias, line 14: Still digesting the code but it seems that there is an assumption that all the switch.handle happen in the beginning and so you are locking the list of protocols we support.

src/stats/index.js
```diff
@@ -0,0 +1,88 @@
...
+  return {
+    stop: stop,
+    global: globalStats,
+    peers: () => Array.from(peerStats.keys()),
+    forPeer: (peerId) => peerStats.get(peerId),
+    transports: () => Array.from(transportStats.keys()),
+    forTransport: (transport) => transportStats.get(transport),
+    protocols: () => Array.from(protocolStats.keys()),
+    forProtocol: (protocol) => protocolStats.get(protocol)
+  }
```
daviddias: How does one "clear" previous stats from a peer?

test/stats.node.js
```diff
@@ -0,0 +1,205 @@
...
+        const intervals = [60000, 300000, 900000]
+        intervals.forEach((interval) => {
+          const average = ma.dataReceived[interval].movingAverage()
+          expect(average).to.be.above(0).below(1)
+        })
+      })
+      teardown(switches, done)
+    })
+  })
+})
```
daviddias: We need tests for:

- [ ] Clearing out previous stats from a peer (Rep mechanisms will need a reset)
- [ ] Check what happens when a peer disconnects and connects again
- [ ] Check when a peer disconnects and connects again from a different transport

daviddias (changes requested):
@pgte really solid work here (and in record time!) 👏🏽Made some comments in what I think is a regression (protocolMuxing) and request for more tests.

src/protocol-muxer.js
pgte: The protocols is an object, which can get mutated later as the user adds more protocols.

src/connection.js
pgte: It's the exact same behaviour as before, with the exception we're tracking the protocol tag for accounting purposes..

test/stats.node.js
pgte: @diasdavid Currently, peer disconnections imply losing the stats for that peer.
If you think that it's important to keep them around, we should impose a time limit so that I eventually becomes garbage collected...
@diasdavid thoughts?

README.md
pgte: @diasdavid can do this, but will be a breaking API change..

pgte: Wait, I see that a peerBook is always required. The only optional argument should be `options`.

src/connection.js
pgte: I would, but `switch` is a reserved JS keyword.. ;)
Switching to `swtch`, in lack of a better replacement…

test/stats.node.js
daviddias: Perhaps tracking the sessions and having them in a LRU cache?

src/protocol-muxer.js
daviddias: Let's not get https://github.com/libp2p/js-libp2p-switch/issues/239 even harder to fix :)

pgte: @diasdavid again, this was the same behaviour as before: `switch.protocols` also contained a mutable object with the protocol handlers... I'm just storing the protocols on that closure, but it's still a JS object that will get mutated, same as before..

test/stats.node.js
pgte: @diasdavid good call, but I'm not going to put all of the sessions there, only once a peer disconnects..

pgte:
I still have one bug to squash that I discovered this morning:
The base protocol (tcp, wbrtc, etc.) is not being accounted for because the peerInfo never resolves on those connections. I'm still trying to track this down..

pgte:
last described issue should be fixed by https://github.com/libp2p/js-libp2p-secio/pull/98

pgte:
@diasdavid everything you asked for (apart from the protocolMuxing regression you described, which I think is not a regression, I argued above) should be done. :)

src/protocol-muxer.js
daviddias: Got it. Makes sense now thanks for explaining :)

hacdias:
Hello @pgte! 😄 When is this PR expected to be finished?

pgte:
I think I’m all done here.

daviddias (approved):
@pgte this looks great!! :) Thanks for kicking ass!

PR #245, opened 2018-02-28
[WIP] feat: handle stream muxer errors

dryajov:
I'm not sure why we weren't handling stream muxer errors here at all, but if there is a good reason and the error handling changes are not needed we can revert them. The really important change is in libp2p-mplex (https://github.com/libp2p/js-libp2p-mplex/pull/75) that prevents the throw from crashing the entire node.


dryajov:
Requires - https://github.com/libp2p/js-libp2p-mplex/pull/75

dryajov:
I'm not sure why we weren't handling stream muxer errors here at all, but if there is a good reason and the error handling changes are not needed we can revert them. The really important change is in libp2p-mplex (https://github.com/libp2p/js-libp2p-mplex/pull/75) that prevents the throw from crashing the entire node.

test/stream-muxers.node.js
```diff
@@ -22,14 +22,15 @@ const Switch = require('../src')
 describe('Stream Multiplexing', () => {
   [
     multiplex,
-    spdy
+    spdy // TODO: do we still support this?
```
daviddias: Of course we still support. The whole design of libp2p is to support multiple building blocks.

test/stream-muxers.node.js
```diff
@@ -149,5 +150,22 @@ describe('Stream Multiplexing', () => {
         }, 500)
       })
     })
+
+    it('should fail graciously on dead stream mux', function (done) {
+      if (sm.multicodec === '/spdy/3.1.0') {
+        this.skip()
+      }
```
daviddias: No skipping for SPDY

src/dial.js
```diff
@@ -208,7 +213,13 @@ function dial (swtch) {
...
 
     function openConnInMuxedConn (muxer, cb) {
-      cb(muxer.newStream())
+      muxer.newStream((err, conn) => {
+        if (err) {
+          return cb(err)
+        }
+
+        cb(null, conn)
+      })
```
daviddias: 👍

src/dial.js
```diff
@@ -208,7 +213,13 @@ function dial (swtch) {
     }
 
     function openConnInMuxedConn (muxer, cb) {
-      cb(muxer.newStream())
+      muxer.newStream((err, conn) => {
```
daviddias: This asking to change the Stream Muxer interface. https://github.com/libp2p/interface-stream-muxer needs to be updated.

daviddias (changes requested):

src/dial.js
dignifiedquire: why not just newStream(cb)?

test/stream-muxers.node.js
dryajov: spdy doesn't return any errors when the stream abruptly terminates. We should make the behaviour consistent then. 

What should happen - error on abrupt termination or silent skip?

test/stream-muxers.node.js
dryajov: Skipping because the behaviour is not consistent across muxers - spdy doesn't return any errors on abrupt termination.

src/dial.js
dryajov: Good question - I'm not sure if this is the correct behavior either, this change is related to -https://github.com/libp2p/js-libp2p-mplex/pull/75, which would simply `throw` on error and crash the whole process. Since I return the error now, it seemed logical to handle it here, but I wasn't sure it is the correct approach, specially since other multiplexers (spdy) don't seem to propagate the errors in the same way. 

So the question remains - should we handle the errors in `switch` or just ignore it?

dryajov:
@diasdavid 

I'm still not sure if we need this at all. This change was driven by -https://github.com/libp2p/js-libp2p-mplex/pull/75, which would simply `throw` on error and crash the whole process. Since I return the error now, it seemed logical to handle it here. Other multiplexer however (spdy) don't seem to follow the same behaviour (not even sure if they can).

So the question remains - should we handle the errors in `switch` or just ignore them?

src/dial.js
dryajov: I didn't really change `newStream` interface, it already takes a callback as it is. I reverted this to return a connection as before. This PR just tries to propagate the error if it's returned by the multiplexer - before it would just be ignored.

harrshasri:
@pgte Any update on this thread? When can I expect this issue to be fixed. If anyone can direct me what needs to be done. I can hop on fixing this.

pgte:
@harrshasri I'm not the one assigned to this. You should probably be speaking to @dryajov :)

harrshasri:
@pgte Aah ... Okay I did tag you just because you were requested for reviewing this.

dryajov:
Thanks for bringing this up @harrshasri, this issue needs some clarification for sure! The current state is as follows:

- this PR depends on https://github.com/libp2p/js-libp2p-mplex/pull/75
   - which depends on https://github.com/libp2p/interface-stream-muxer/pull/49
     - which adds tests for the specific case introduced in this PR and in `js-libp2p-mplex`
    -  however, we would also potentially need to changes to `js-libp2p-spdy` to make the behavior consistent across the two multiplexers. 

Also https://github.com/libp2p/js-libp2p-mplex/pull/75 might not be required anymore since https://github.com/libp2p/js-libp2p-mplex/pull/76 is almost here, but untill then the above dependencies need to be resolved.

dryajov:
not relevant anymore

PR #246, opened 2018-03-12
fix: switch observer edge cases

pgte:
For when:

* there is not protocol handler funcion
* peerInfo resolves to `undefined`


pgte:
Should fix CI build on https://github.com/libp2p/js-libp2p/pull/169

daviddias:
Released as v0.37.1

PR #252, opened 2018-04-25
refactor: update files and add jsdocs to improve readability

jacobheun:
In preparation for the changes that will need to occur to get the Private Network implemented, libp2p/js-libp2p#163, I started doing some refactoring to the Switch in order to improve readability and update jsdocs throughout the repo. I think this refactor will also make it easier to turn Switch into a state machine libp2p/js-libp2p-switch#24.

There is still some more work I would like to do with this, but I wanted to get visibility to this sooner rather than later. As someone who was really diving into this repo for the first time, it was difficult to get up to speed on everything that's going on, especially in dial.js. I think this helps provide more clarity to that and should make it easier for others to contribute in the future.

I'd love some feedback on these changes.

**Update**: Okay, I've done some more refactoring, added additional jsdocs and comments and added in some tests for the get-peer-info.js file.  I think this is in a good position to get merged in. I avoided making any significant logic changes for this so that it's a pure refactor. I'd like to build the state machine and private network efforts of of this if possible.

## Verified against
The refactor will be checked against the following repos by locally linking switch to the repo's and running their test suites.

- [x] libp2p/js-libp2p
- [x] ipfs/js-ipfs
- [x] ipfs/interop

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -143,8 +143,6 @@ class ConnectionManager {
         Object.assign(config, { hop: { enabled: false, active: false } })
       }
 
-      // TODO: (dryajov) should we enable circuit listener and
-      // dialer by default?
       this.switch.transport.add(Circuit.tag, new Circuit(this.switch, config))
     }
   }
--- a/src/connection.js
+++ b/src/connection.js
@@ -12,121 +12,155 @@ const Circuit = require('libp2p-circuit')
 
 const plaintext = require('./plaintext')
 
-module.exports = function connection (swtch) {
-  return {
-    addUpgrade () {},
+/**
+ * Contains methods for binding handlers to the Switch
+ * in order to better manage its connections.
+ */
+class ConnectionManager {
+  constructor (_switch) {
+    this.switch = _switch
+  }
 
-    addStreamMuxer (muxer) {
-      // for dialing
-      swtch.muxers[muxer.multicodec] = muxer
+  /**
+   * Adds a listener for the given `muxer` and creates a handler for it
+   * leveraging the Switch.protocolMuxer handler factory
+   *
+   * @param {Muxer} muxer
+   * @returns {void}
+   */
+  addStreamMuxer (muxer) {
+    // for dialing
+    this.switch.muxers[muxer.multicodec] = muxer
 
-      // for listening
-      swtch.handle(muxer.multicodec, (protocol, conn) => {
-        const muxedConn = muxer.listener(conn)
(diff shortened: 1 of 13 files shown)


src/connection.js
```diff
@@ -12,121 +12,155 @@ const Circuit = require('libp2p-circuit')
...
-      swtch.handle(tag, (protocol, conn) => {
-        const myId = swtch._peerInfo.id
-        const secure = encrypt(myId, conn, undefined, () => {
-          swtch.protocolMuxer(null)(secure)
-        })
-      })
-
-      swtch.crypto = {tag, encrypt}
+      // TODO: (dryajov) should we enable circuit listener and
+      // dialer by default?
```
daviddias: We should not as it is an option

daviddias (dismissed):
There is a lot of great work here, @jacobheun 🌟

Given how sensitive this code is, let's be extra careful and make sure that tests on js-libp2p, js-ipfs and ipfs/interop pass with these changes as well. Could you make sure that happens, @jacobheun ?

daviddias:
@jacobheun I got some issues yesterday with deps being out of date on the switch->libp2p->js-ipfs that got fixed just by bubbling up updates. Rebasing master on this branch might solve the issues you saw before.

jacobheun:
@diasdavid thanks, I will take a look. I found a small issue yesterday when testing with libp2p but was able to resolve that without much issue. js-ipfs is the most difficult one to verify due to the flakey tests. I'll do some more validation on that today as well as checking the rebase against interop and libp2p again.

jacobheun:
@diasdavid I have run tests against interop, libp2p and js-ipfs (via libp2p). After the fix I made yesterday and rebasing with master, I haven't encountered any issues linking locally. We should be good to merge.

src/connection.js
```diff
@@ -12,121 +12,153 @@ const Circuit = require('libp2p-circuit')
...
+          }
+          const b58Str = peerInfo.id.toB58String()
+
+          this.switch.muxedConns[b58Str] = { muxer: muxedConn }
+
+          if (peerInfo.multiaddrs.size > 0) {
+            // with incomming conn and through identify, going to pick one
+            // of the available multiaddrs from the other peer as the one
+            // I'm connected to as we really can't be sure at the moment
+            // TODO add this consideration to the connection abstraction!
```
daviddias, line 82: Potentially this could try the getObserverdAddrs first.

daviddias (approved):

PR #254, opened 2018-05-09
fix deps

dryajov:
NOTE: Im using libp2p-mplex as a git dep, so that we can test it on CI. This is temporary and should be changed once libp2p-mplex is released.


dryajov:
These PRs are only needed to ensure CI runs correctly with the new implementation. It is not supposed to get merged just yet, it need pull-mplex and libp2p-mplex published first. Please don't change them.

test/stats.node.js
```diff
@@ -10,7 +10,7 @@ const each = require('async/each')
 const map = require('async/map')
 const series = require('async/series')
 const TCP = require('libp2p-tcp')
-const multiplex = require('libp2p-multiplex')
+const multiplex = require('libp2p-mplex')
```
daviddias, line 13: @dryajov this was a legit issue that fixes it.

PR #256, opened 2018-05-22
fix: libp2p/js-libp2p#189 Prevent self-dial

mkg20001:


jacobheun (approved):
Looks good, thanks for adding the tests!

PR #260, opened 2018-05-29
fix: prevent undefined error during a mutual hangup

jacobheun:
This resolves an issue where if connected nodes hang up at the same time, such as during testing, a muxedConnection could be deleted from the switch at the same time as it's async iteration was occurring to end those connections. This would result in an undefined error.

This PR ensures we don't attempt to close connections that have been hung up on.

You can see the passing CI build for js-libp2p that uses this branch here: https://ci.ipfs.team/blue/organizations/jenkins/libp2p%2Fjs-libp2p/detail/chore%2Fupdate-switch/4/pipeline

connects to https://github.com/libp2p/js-libp2p/issues/198

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -110,6 +110,11 @@ class Switch extends EE {
     this.stats.stop()
     series([
       (cb) => each(this.muxedConns, (conn, cb) => {
+        // If the connection was destroyed while we are hanging up, continue
+        if (!conn) {
+          return cb()
+        }
+
         conn.muxer.end((err) => {
           // If OK things are fine, and someone just shut down
           if (err && err.message !== 'Fatal error: OK') {
--- a/src/dial.js
+++ b/src/dial.js
@@ -267,8 +267,9 @@ class Dialer {
           }
 
           const muxedConn = this.switch.muxers[key].dialer(conn)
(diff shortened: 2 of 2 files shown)


src/index.js
```diff
@@ -110,6 +110,11 @@ class Switch extends EE {
     this.stats.stop()
     series([
       (cb) => each(this.muxedConns, (conn, cb) => {
+        // If the connection was destroyed while we are hanging up, continue
+        if (conn === undefined) {
```
dryajov: Should this check be so restrictive? Perhaps an `if (conn) {...}` should do?

dryajov (commented):
LGTM, other than the comment I left.

src/index.js
jacobheun: I originally had a more generic check `if (!conn)`, but changed it to undefined as there shouldn't be a situation where conn is something other than undefined or an object with the muxer property.  The only way it would be something other than those two options is if deletions of the muxers changes internally, or if they are overridden by another library.

I suppose the tiny performance benefit during a swarm stop isn't worth the longer term resilience of the generic check. I can update this.

jacobheun: Update pushed, this should be good to go.

PR #261, opened 2018-05-30
fix: return on call to nextMuxer

alanshaw:
When the call to multistream.Dialer.select is unsuccessful, call nextMuxer to try select the next one in the list but do not continue executing callback afterwards.


jacobheun (approved):
Good catch! Looks good.

PR #262, opened 2018-05-30
fix: drop connection when stream ends unexpectedly

### LINKED ISSUE #235: Error "already piped"
`js-ipfs 0.4.7` was running on a windows machine for a few hours and then raised the following error.

```
C:\Users\Owner\Documents\GitHub\js-ipfs>node src\cli\bin.js daemon
Initializing daemon...
Swarm listening on /p2p-circuit/ipfs/QmYHExHtTzFEjdwyQcJkD9gaKKgfwvGswoFb4simrJn5Q6
Swarm listening on /p2p-circuit/ip4/0.0.0.0/tcp/4002/ipfs/QmYHExHtTzFEjdwyQcJkD9gaKKgfwvGswoFb4simrJn5Q6
Swarm listening on /p2p-circuit/ip4/127.0.0.1/tcp/4003/ws/ipfs/QmYHExHtTzFEjdwyQcJkD9gaKKgfwvGswoFb4simrJn5Q6
Swarm listening on /ip4/127.0.0.1/tcp/4003/ws/ipfs/QmYHExHtTzFEjdwyQcJkD9gaKKgfwvGswoFb4simrJn5Q6
Swarm listening on /ip4/192.168.178.21/tcp/4002/ipfs/QmYHExHtTzFEjdwyQcJkD9gaKKgfwvGswoFb4simrJn5Q6
Swarm listening on /ip4/127.0.0.1/tcp/4002/ipfs/QmYHExHtTzFEjdwyQcJkD9gaKKgfwvGswoFb4simrJn5Q6
API is listening on: /ip4/127.0.0.1/tcp/5002
Gateway (readonly) is listening on: /ip4/127.0.0.1/tcp/9090
Daemon is ready
C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\pull-pair\index.js:11
      throw new Error('already piped')
      ^

Error: already piped
    at sink (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\pull-pair\index.js:11:13)
    at consume (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\pull-defer\sink.js:7:17)
    at Connection.consume (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\pull-defer\sink.js:7:17)
    at pull (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\pull-stream\pull.js:43:9)
    at Dialer.handle (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\multistream-select\src\dialer\index.js:47:5)
    at attemptMuxerUpgrade (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\libp2p-swarm\src\dial.js:159:10)
    at gotWarmedUpConn (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\libp2p-swarm\src\dial.js:53:7)
    at Swarm.dial (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\libp2p-swarm\src\dial.js:40:9)
    at _getPeerInfo (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\libp2p\src\index.js:265:18)
    at setImmediate (C:\Users\Owner\Documents\GitHub\js-ipfs\node_modules\libp2p\src\index.js:324:24)
```

### LINKED ISSUE #1366: Revenge of the `Missing peerId` and its sidekick, `already piped`
- **Version**: 0.28.1
- **Platform**: Firefox 60.0.1
- **Subsystem**: Unknown

#### Description:

This is related to #1221 and probably to the issues @shessenauer reported there a few days ago. I am opening a new issue since this is in the browser, and since I am using js-ipfs v0.28.1, which has a fix for #1221. **WRONG! This version predates that fix! It looks like the fix hasn't made it to a release yet.**

I am working on https://interfect.github.io/ipfstunes/

I am getting this in my browser console:

```
uncaught exception: AssertionError: Missing peerId. Use Peer.create(cb) to create one
```

And also this (with a stacktrace):

```
Error: already piped
```

It only happens when I give one node a reason to talk to another, by adding some songs in one browser and then opening the URL with fragment identifier in another browser and trying to play them.

I'm seeing the messages on both the sending and receiving browser. I'm also experiencing very slow file transfer between the browsers, and I suspect the two phenomena might be related.

Here's a stack trace for the `already piped` error:
```
Error: already piped
ipfs.js:51632:13
sink
https://interfect.github.io/ipfstunes/lib/ipfs.js:51632:13
consume
https://interfect.github.io/ipfstunes/lib/ipfs.js:76086:17
consume
https://interfect.github.io/ipfstunes/lib/ipfs.js:76086:17
consume
https://interfect.github.io/ipfstunes/lib/ipfs.js:76086:17
pull
https://interfect.github.io/ipfstunes/lib/ipfs.js:69916:7
handle
https://interfect.github.io/ipfstunes/lib/ipfs.js:204460:5
attemptMuxerUpgrade
https://interfect.github.io/ipfstunes/lib/ipfs.js:208185:7
gotWarmedUpConn
https://interfect.github.io/ipfstunes/lib/ipfs.js:208082:7
Ipfs</dial/<
https://interfect.github.io/ipfstunes/lib/ipfs.js:208068:9
Ipfs</dial/<
https://interfect.github.io/ipfstunes/lib/ipfs.js:203766:7
Ipfs</</module.exports/</<
https://interfect.github.io/ipfstunes/lib/ipfs.js:209951:24
run
https://interfect.github.io/ipfstunes/lib/ipfs.js:68437:13
runIfPresent
https://interfect.github.io/ipfstunes/lib/ipfs.js:68466:21
onGlobalMessage
https://interfect.github.io/ipfstunes/lib/ipfs.js:68506:17
```

#### Steps to reproduce the error:

* Have two browsers
* Open https://interfect.github.io/ipfstunes/ in one
* Hit Browse and import some MP3 files
* Wait for the import to finish
* Copy the resulting URL (with fragment identifier) from the first browser to the second
* Wait for the second browser to find the first and successfully download the song database
* Hit Search in the second browser (to search for the empty string)
* Click on some results in the list
* Click the play button
* Wait for the second browser to (slowly!) download and play the songs you selected from the first browser
* The reported error(s) should appear from time to time in both browser consoles

olizilla:
Pull streams pass true in the error position when the sream ends.
In https://github.com/multiformats/js-multistream-select/blob/5b19358b91850b528b3f93babd60d63ddcf56a99/src/select.js#L18-L21
...we're getting lots of instances of pull-length-prefixed stream
erroring early with `true` and it's passed back up to the dialer
in https://github.com/libp2p/js-libp2p-switch/blob/fef2d11850379a4720bb9c736236a81a067dc901/src/dial.js#L238-L241

The `_createMuxedConnection` contains an assumption that any error
that occurs when trying `_attemptMuxerUpgrade` is ok, and keeps the
relveant baseConnecton in the cache. If the pull-stream has ended
unexpectedly then keeping the connection arround starts causing
the "already piped" errors when we try and use the it later.

This PR adds a guard to avoid putting the connection back into the
cache if the stream has ended.

There is related work in an old PR to add a check for exactly this issue in
pull-length-prefixed https://github.com/dignifiedquire/pull-length-prefixed/pull/8
...but it's still open, so this PR adds a check for true in
the error position at the site where the "already piped" errors
were appearing. Once the PR on pull-length-prefixed is merged this
check can be removed. It's not ideal to have it in this code as it
is far removed from the source, but it fixes the issue for now.

Arguably anywhere that `msDialer.handle` is called should do the
same check, but we're not seeing this error occur anywhere else so
to keep this PR small, I've left it as the minimal changeset to
fix the issue.

Of note, we had to add '/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star'
to the swarm config to trigger the "already piped" errors. There
is a minimal test app here https://github.com/tableflip/js-ipfs-already-piped-error

Manual testing shows ~50 streams fail in the first 2 mins of
running a node, and then things stabalise with ~90 active muxed
connections after that.

Fixes #235
Fixes https://github.com/ipfs/js-ipfs/issues/1366
See https://github.com/dignifiedquire/pull-length-prefixed/pull/8

License: MIT
Signed-off-by: Oli Evans <oli@tableflip.io>

### DIFF
--- a/src/dial.js
+++ b/src/dial.js
@@ -11,6 +11,53 @@ const log = debug('libp2p:switch:dial')
 
 const getPeerInfo = require('./get-peer-info')
 const observeConnection = require('./observe-connection')
+const UNEXPECTED_END = 'Unexpected end of input from reader.'
+
+/**
+ * Uses the given MultistreamDialer to select the protocol matching the given key
+ *
+ * A helper method to catch errors from pull streams ending unexpectedly
+ * Needed until https://github.com/dignifiedquire/pull-length-prefixed/pull/8 is merged.
+ *
+ * @param {MultistreamDialer} msDialer a multistream.Dialer
+ * @param {string} key The key type to select
+ * @param {function(Error)} callback Used for standard async flow
+ * @param {function(Error)} abort A callback to be used for ending the connection outright
+ * @returns {void}
+ */
+function selectSafe (msDialer, key, callback, abort) {
+  msDialer.select(key, (err, conn) => {
+    if (err === true) {
+      return abort(new Error(UNEXPECTED_END))
+    }
+
+    callback(err, conn)
+  })
+}
+
+/**
+ * Uses the given MultistreamDialer to handle the given connection
+ *
+ * A helper method to catch errors from pull streams ending unexpectedly
+ * Needed until https://github.com/dignifiedquire/pull-length-prefixed/pull/8 is merged
+ *
+ * @param {MultistreamDialer} msDialer
+ * @param {Connection} connection The connection to handle
+ * @param {function(Error)} callback Used for standard async flow
+ * @param {function(Error)} abort A callback to be used for ending the connection outright
+ * @returns {void}
+ */
+function handleSafe (msDialer, connection, callback, abort) {
+  msDialer.handle(connection, (err) => {
+    // Repackage errors from pull-streams ending unexpectedly.
+    // Needed until https://github.com/dignifiedquire/pull-length-prefixed/pull/8 is merged.
+    if (err === true) {
+      return abort(new Error(UNEXPECTED_END))
+    }
+
+    callback(err)
+  })
+}
 
 /**
  * Manages dialing to another peer, including muxer upgrades
@@ -55,6 +102,11 @@ class Dialer {
         cb(null)
       }
     ], (err, connection) => {
+      if ((err && err.message === UNEXPECTED_END) || err === true) {
+        log('Connection dropped for %s', this.peerInfo.id.toB58String())
+        return this.callback(null, null)
+      }
+
       this.callback(err, connection)
     })
 
@@ -160,13 +212,6 @@ class Dialer {
 
     connection.setPeerInfo(this.peerInfo)
     this._attemptMuxerUpgrade(connection, b58Id, (err, muxer) => {
-      // The underlying stream closed unexpectedly, so drop the connection.
-      // Fixes https://github.com/libp2p/js-libp2p-switch/issues/235
-      if (err && err.message === 'Unexpected end of input from reader.') {
-        log('Connection dropped for %s', b58Id)
-        return callback(null, null)
-      }
-
       if (err && !this.protocol) {
         this.switch.conns[b58Id] = connection
         return callback(null, null)
@@ -178,7 +223,7 @@ class Dialer {
       }
 
       callback(null, muxer)
-    })
+    }, callback)
   }
 
   /**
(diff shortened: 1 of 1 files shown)


olizilla:
This should have a test, but it's late and I wanted to surface that there is a fix, and get some advice on how to go about testing it.

Also, this PR removes an un-needed waterfall with only a single task in it from `_createMuxedConnection`, because @alanshaw wanted it so.

jacobheun:
This looks good. I will think about the testing options a bit more, as it's late here as well. I am running your test app locally and that looks good. My initial thoughts are that this is going to be difficult to test, without significant changes, in switch's current state. I am leaning towards releasing this without them as the changes and risk are small, and there will be a clear reduction in already piped errors.

The work to convert the underlying connections and switch itself into state machines, https://github.com/libp2p/js-libp2p-switch/issues/24, should make this much easier to catch, handle and test these types of errors.

olizilla:
Agreed. It'd be more valuable to refactor js-multistream-select and add a smoke test for js-ipfs that just does what the demo app does... let ipfs run for a few minutes in a real browser and see if any errors occur. It's not a tidy unit test, but it would have picked this up before the 0.29 release.

jacobheun:
@olizilla after running your example more I ran into the error again. After looking at the stack trace I think this is indeed due to the other multistream dialers not handling the `true` error.

I made an update to the code locally and am testing it against your example. It adds two helper methods `handleSafe` and `selectSafe` that check for the error, and those are used in place of the existing dialer.handle and dialer.select logic. I'm doing some final testing now and can push that up shortly for you to take a look at.

jacobheun:
@olizilla here is the commit I am testing against, https://github.com/libp2p/js-libp2p-switch/commit/af27db9c065578f6b528e31f27a49ac5d2e90d0a. I didn't want to push directly to your branch, but I can cheery pick that over if it looks good to you. 

There are a bit more changes, but I haven't encountered the error yet and am getting quite a few `Connection dropped` logs. Looking at the logs it's also interesting to see there appear to be 4 particular addresses that are problematic.

olizilla:
@jacobheun feel free to push to this branch. I didn't see errors occurring from the other usages, I'd have done similar to what you have.

I think keeping the "ok errors" separate from the "error errors" is a good move and a reasonable fix here for now, so I'm all +1 on those changes.

More generally, I'm not familiar enough with the purpose of each layer to say but it feels like the the number of errors that can be safely ignored would be finite, so we could switch things so that only a known set of errors are tested against to suggest that the connection is still ok even though a transition failed. That or MultistreamDialer should take responsibility for signalling the difference between "temporary errors" and "failed connections" to the caller. These are just musings at this point, I'm fine with fixing this at the switch layer for now, but we should follow up with @diasdavid about what (if anything) should happen to the api for MultistreamDialer to simplify this situation for calling code.

jacobheun:
Yeah, I think this is something that needs to be solved in the `multistream`. I would expect handle and select to catch and return errors in the callback, rather than those errors being thrown. 

While this fix should help mitigate the error, we're likely to still encounter it until the underlying problem is resolved.

jacobheun (approved):

PR #265, opened 2018-06-18
fix: better websocket errors

### LINKED ISSUE #1326: Websocket connection errors on startup should be clearer and not crash IPFS
- **Version**: "ipfs": "^0.28.2"
- **Platform**: Linux on AWS EC2 (Dockerized API)
- **Subsystem**: /ipfs/src/core/boot.js or p2plib?

#### Description:
Upon start of the application, it starts normally - then as soon as the IPFS js node gets initialized, the application throws a critical error with the following: 
```
App starting..

events.js:165
      throw er; // Unhandled 'error' event
      ^

Error: websocket error
    at WS.Transport.onError (/path/to/application/node_modules/engine.io-client/lib/transport.js:64:13)
    at WebSocket.ws.onerror (/path/to/application/node_modules/engine.io-client/lib/transports/websocket.js:150:10)
    at WebSocket.onError (/path/to/application/node_modules/ws/lib/EventTarget.js:109:16)
    at WebSocket.emit (events.js:180:13)
    at WebSocket.finalize (/path/to/application/node_modules/ws/lib/WebSocket.js:182:41)
    at ClientRequest._req.on (/path/to/application/node_modules/ws/lib/WebSocket.js:653:12)
    at ClientRequest.emit (events.js:180:13)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:539:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:117:17)
    at TLSSocket.socketOnData (_http_client.js:444:20)
Emitted 'error' event at:
    at done (/path/to/application/node_modules/ipfs/src/core/boot.js:58:19)
    at /path/to/application/node_modules/async/internal/parallel.js:39:9
    at /path/to/application/node_modules/async/internal/once.js:12:16
    at iterateeCallback (/path/to/application/node_modules/async/internal/eachOfLimit.js:44:17)
    at /path/to/application/node_modules/async/internal/onlyOnce.js:12:16
    at /path/to/application/node_modules/async/internal/parallel.js:36:13
    at done (/path/to/application/node_modules/ipfs/src/core/components/start.js:15:16)
    at series (/path/to/application/node_moduless/ipfs/src/core/components/start.js:39:25)
    at /path/to/application/node_modules/async/internal/parallel.js:39:9
    at /path/to/application/node_modules/async/internal/once.js:12:16

```

In the configuration, I have a node.js application that uses IPFS-js. I have the application dockerized as well as a full ipfs dockerized node that websockets as a peer to the IPFS-js app. This works perfectly fine and I have had no problem until last night when every single one of my environments started throwing the same error out of nowhere. Now none of my team can spin up the api because it has the same issue. One odd thing is that in the gap of time before it crashes, we can spam the API with a curl and get back the IPFS results for a quick second, otherwise it fails. I have confirmed it is not a websocket issue on my end and it is infact something with the IPFS-js package in the nodeJS application. Guidance is much appreciated.

#### Steps to reproduce the error:

1. npm install ipfs
2. set the options for websockets 
```
let options = {
  config: {
    Addresses: {
      Swarm: [
        // '/dns4/wrtc-star.discovery.libp2p.io/tcp/443/wss/p2p-webrtc-star'
        '/dns4/ws-star.discovery.libp2p.io/tcp/443/wss/p2p-websocket-star'
      ]
    }
  }
}
```
3. initialize the node 
```const ipfsNode = new IPFS(options)```
```
4.  Connect to swarm
ipfsNode.on('ready', () => {
  // Your node is now ready to use \o/
  console.log('IPFS Online Status: ', ipfsNode.isOnline())
  ipfsNode.swarm.connect(ipfsPeer, (err, result) => {
    console.log('connecting to peers: ', result)
    ipfsNode.swarm.peers((err, peerCount) => {
      console.log('There are this many peers: ', peerCount)
    })
  })
  ```

fsdiogo:
This updates the error when a failed WebSocket connection occurs.

A first step to fix https://github.com/ipfs/js-ipfs/issues/1326.

As stated in that issue, a failed connection should not crash the daemon, we should try to reconnect from time to time.


src/transport.js
```diff
@@ -128,6 +129,15 @@ class TransportManager {
 
     parallel(createListeners, (err) => {
       if (err) {
+        // Error connecting to WebSocket
+        if (err.description && err.description.code === 'ENOTFOUND') {
+          const hostname = err.description.hostname
+
+          err = Object.assign(new Error(`WebSocket connection failed on ${hostname}`), {
```
dryajov, line 136: I don't think this pertains in the switch - which is transport agnostic. If anything, this should be emitted by the transport itself.

dryajov (changes requested):

src/transport.js
fsdiogo: Makes sense. I should probably close this PR and put this into [js-libp2p-websocket-star](https://github.com/libp2p/js-libp2p-websocket-star) then?

jacobheun: Since the error is specific to websockets it makes sense for that to exist in js-libp2p-websocket-star and not here. I think there are two things to take into account here. 

1. This PR looks like you're just trying to get more meaningful errors logged/thrown when things break to get better information to developers. If that's the case, creating that error in websocket-star and letting it propagate through libp2p-switch to the user should be sufficient for the short term.

2. The larger issue this brings up is better error messaging and handling in general. Right now, a lot of errors coming out of switch via the transports aren't terribly helpful and can be really hard to debug. I think what I'd like to  see happen for this is to create the generic, extendable error types in https://github.com/libp2p/interface-connection and then have each transport catch and add meaningful metadata to those, before passing them on. This would enable switch, and other libp2p layers to handle the generic errors and take action accordingly, while giving users more detailed metadata from specific transports. If the error is meant to be fatal we can handle that appropriately. When the more common scenario of a bad connection occurs, we can simply terminate that connection and move on. Combining this with a move of switch to a state machine would be really helpful in making this a more resilient, debuggable system. cc: @diasdavid

fsdiogo: Thanks for the input @dryajov and @jacobheun. I'll close this one.

@jacobheun I'm going to open a more generic issue about error handling and daemon stability and it would be cool to have your second point there for an open discussion.

PR #268, opened 2018-07-09
Fix compatibility with Parcel bundler

antimatter15:
I was trying to run libp2p in the browser with the Parcel bundler. However it would always fail at 'js-libp2p-switch/stats/stat.js' with the error that `Big is not a function`. It appears that `Big` gets imported as an object `{ default: [Function], Big: [Function], __esModule: true }` but the code is trying to run it directly. 

Big.js is now an ES6 module which exports "default" and "Big" fields (which are both the same function).

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
   },
   "dependencies": {
     "async": "^2.6.0",
-    "big.js": "^5.0.3",
+    "big.js": "^5.1.2",
     "debug": "^3.1.0",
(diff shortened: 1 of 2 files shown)


mkg20001 (approved):

jacobheun:
Thanks @antimatter15! This looks good. I rebased the commits into master to fix the commitlint task in jenkins. The changes are available in the new release, v0.40.5.

PR #270, opened 2018-07-09
[bugfix] stats - observer expects protocolTag

kumavis:
bug was introduced here when `protocol` variable was renamed, but not all usages were renamed
https://github.com/libp2p/js-libp2p-switch/commit/6265ec639ac7c42e81221320152fde36b83b2dd2

### DIFF
--- a/src/protocol-muxer.js
+++ b/src/protocol-muxer.js
@@ -18,7 +18,7 @@ module.exports = function protocolMuxer (protocols, observer) {
         if (protocol) {
           const handlerFunc = protocol && protocol.handlerFunc
           if (handlerFunc) {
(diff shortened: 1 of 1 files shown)


dryajov (approved):
LGTM

kumavis:
force pushed with proper semantic git commit message

kumavis:
some failure
```
1) Switch (WebSockets)
       close:
     Timeout of 5000ms exceeded. For async tests and hooks, ensure "done()" is called; if returning a Promise, ensure it resolves.
```

kumavis:
possibly why this was not caught
https://github.com/libp2p/js-libp2p-switch/issues/269

jacobheun (approved):
Great catch, thank you!

Issue #269, opened 2018-07-09
Stats tests are not being run


jacobheun:
Since the stats test file is not a `.spec.js` file, they are not run automatically so it must be included by the browser / node test runner files. We should do an evaluation of the test suite to make sure everything is being run and in the proper environments.

dryajov:
The current stats tests relay on the `tcp` transport which isn't going to work in the browser, so we should either rework the tests to run in both node and browser, or create a separate set of tests for browser environments. For now, I think we should enable them by requiring them in the `test/node.js` file.

PR #271, opened 2018-07-09
fix: re-enable stats tests in node

dryajov:
Addresses - https://github.com/libp2p/js-libp2p-switch/issues/269#issuecomment-403574647


jacobheun (approved):
LGTM! I'll leave #269 open until we resolve the broswer tests for stats.

PR #272, opened 2018-07-12
Fix/dial proto stats

dryajov:
Protocol stats were not being counted correctly for dialer multistreams. Also, multiplexed connections where also not being tracked on multiplex creation, but where being counted as multistream protos on the listener end.

### DIFF
--- a/src/dial.js
+++ b/src/dial.js
@@ -218,6 +218,7 @@ class Dialer {
       }
 
       if (err) {
+        log('muxer upgrade failed with error', err)
         // couldn't upgrade to Muxer, it is ok, use the existing connection
         return callback(null, connection)
       }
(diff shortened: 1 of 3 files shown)


src/dial.js
```diff
@@ -404,7 +408,8 @@ class Dialer {
         return callback(err)
       }
 
-      selectSafe(msDialer, this.protocol, (err, conn) => {
+      selectSafe(msDialer, this.protocol, (err, _conn) => {
+        const conn = observeConnection(null, this.protocol, _conn, this.switch.observer)
```
kumavis: prolly need to abort on error here

kumavis:
this solves the mysterious missing stats bug we were seeing 👍

src/dial.js
dryajov: Actually, it seems like the error just gets propagated through the `callback` in the following line.

kumavis: im mean that the `observeConnection` call might break if we pass in an undefined `_conn` in the event of an `err`

kumavis: the ol `if (err) return cb(err)` abort statement

dryajov: you are right, that part should be more standard anyways.

jacobheun (approved):
Nice catch! The changes look good

PR #274, opened 2018-07-26
ws error pull swarm

Harzu:
Hello, my name is Ilya Shvyryalkin. I am a developer at DCLabs (Dao Casino). When developing a messaging using ipfs-pubsub-room, we came across the following problem: I declared that I want to connect, and it's logical that if one falls off, then I need to connect to the next one, but it does not happen and the error falls out. The study of this problem led me to this module - libp2p. Fixed this issue by removing it from the pool. A few addresses that throws an error and restarts the start method. I also write the index and the address of the wrong server in the error so that it can be processed later on the client side. Tests drove out, everything works and our problem solved it. It would be very cool if you accepted these changes or wrote how you can fix this problem more elegantly.

### DIFF
--- a/src/connection.js
+++ b/src/connection.js
@@ -79,7 +79,7 @@ class ConnectionManager {
             // with incomming conn and through identify, going to pick one
             // of the available multiaddrs from the other peer as the one
             // I'm connected to as we really can't be sure at the moment
-            // TODO add this consideration to the connection abstraction!
+            // This is TODO: add this consideration to the connection abstraction!
             peerInfo.connect(peerInfo.multiaddrs.toArray()[0])
           } else {
             // for the case of websockets in the browser, where peers have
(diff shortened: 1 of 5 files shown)


jacobheun:
Hi @Harzu! I'm interested in better understanding your use case to better understand the needed solution. Currently the switch will check all of the addresses and only listen with the transports that support those addresses. It sounds like you are encountering an issue where on listening, a given address can't be listened on. 

This should indicate that either the address is bad, or the transport cannot listen on that address if it is valid. To me, this would suggest a need to fix the configuration of the node to not include the transport that can't listen on that nodes setup. Is updating the configuration not an option for your use case?

jacobheun:
Closing this due to inactivity, please feel free to open a new issue or PR if you're still having issues.

PR #275, opened 2018-08-14
test: add sample network circuit relay tests

jacobheun:
Added some additional tests around circuit relays to make sure that various dials are working properly and that peer connection events are being sent for peers connecting via the relay (https://github.com/libp2p/js-libp2p-circuit/issues/26)

### DIFF
--- a/test/circuit-relay.node.js
+++ b/test/circuit-relay.node.js
@@ -14,6 +14,7 @@ const TCP = require('libp2p-tcp')
 const WS = require('libp2p-websockets')
 const multiplex = require('libp2p-mplex')
 const PeerBook = require('peer-book')
+const getPorts = require('portfinder').getPorts
 
 const utils = require('./utils')
 const createInfos = utils.createInfos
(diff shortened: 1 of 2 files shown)


test/circuit-relay.node.js
```diff
@@ -136,4 +137,189 @@ describe(`circuit`, function () {
...
+      expect(err).to.not.exist()
+
+      bootstrapPeer = infos[0]
+      tcpPeer1 = infos[1]
+      tcpPeer2 = infos[2]
+      wsPeer1 = infos[3]
+      wsPeer2 = infos[4]
+
+      // Setup the addresses of our nodes
+      bootstrapPeer.multiaddrs.add('/ip4/127.0.0.1/tcp/9100')
```
dryajov: should use ephemeral ips and ports

dryajov (changes requested):
should use ephemeral ips and ports, other than that 👍

jacobheun:
Ports are now being discovered. I noticed I was also not stopping the swarms after the tests, so that is being done now. Most of the tests in switch are not using ephemeral ports. If we're doing actual connection testing we should migrate those over.

dryajov:
> Most of the tests in switch are not using ephemeral ports.

huh, true - I forgot about #227! But you're not using port `0` either way so I think this will work?

dryajov (approved):

Issue #24, opened 2016-03-10
be a state machine


diasdavid:
I've been debating with myself for some hours now of how to solve a old problem in the most elegant way possible.

It resumes to this:

a) If a connection is dropped (for reasons), a socket will emit that error, if not catched, it will make the program crash. 

b) If a connection, that was upgraded to a stream multiplexer drops, we get the error on the connection + an error emitted in every single multiplexed conn that was created on top of that original connection. 

This is kind of annoying, because connections will drop at all the times we could do something hacky and capture all the errors at least once so that the process doesn't crash. However this doesn't solve the problem of handling a crash on a connection that holds several multiplexed connections between two peers, because the most dumb solution would be for the app or apps (bitswap, peer routing, record store, etc) realising that their connection broke and ask to dial again, making several points in the code to try to dial to the same peer again, and to make things worse,  when you support multiple transports, stream muxers, possible relays, etc, the level of madness scales to infinity.

Not saying that with a very complex piece of code that tests to every single edge case, this wouldn't be possible, but if we could find a elegant solution to describe the conn broke problem, it would be 🌟

noffle:
I don't know enough about the implementations to say if this is the case or not, but one way of looking at upgraded connections is that the connection object represents a connection not *to its underlying destination* but rather *to the multiplexer*. Under this model, a transport-level disconnect becomes a decision made by the multiplexer: either report the error to its sub-streams (but a redial operation would just be an inexpensive signal to the multiplexer asking it to re-establish its connection to the underlying endpoint), or even refuse to pass on the error to the sub-streams (enabling mosh-like persistent connections over unreliable or intermittent transports).

pgte:
Assuming that messages are idempotent, the way I would design the thing would be:

You will have to construct a stream that, underneath, reconnects if it needs to, providing a duplex stream interface to the outside world.
On top of this stream you can then place the multiplexer.

Like this, the multiplexer would never see the underlying connection disrupted (unless a serious error happened, or a maximum number of reconnect attempts was reached, etc.)

The problem with this is that the underlying transport now doesn't guarantee the delivery of messages because you could be reconnecting all the time. Depending on your protocol, this could be hard to manage (do you do request-response, pub/sub, events?). You could have to retry messages, and that would break the abstraction.

Because of this, you would need a transport (and probably also a heartbeat) protocol on top of the reconnect-enabled stream, with buffering.

[...]

PR #278, opened 2018-09-26
Make switch a state machine

### LINKED ISSUE #239: Dialers can request unencrypted protocols via multistream-select.
Not sure if this is an intentional design choice. Dialers can request any registered protocol during the first phase of the connection when the encrypted channel is being selected.

I was able to connect to other peers without encryption by replacing the call to cryptoDial() in dial.js with gotWarmedUpConn(conn).

It doesn't appear there is a way to disallow this on the listener end since the same array of protocols (containing both the crypto protocol and others) is used throughout. 

As a workaround, users could always manually wrap their stream with an encrypted one.

jacobheun:
Getting this pushed up for visibility while I work on cleanup and add some additional tests.

## Goal
This PR updates libp2p-switch to be a state machine, including its connections #24. 

## Updates
* Switch will now emit `start`, `stop`, and `error` events. Callback legacy support is included.
* Incoming connections will now ensure crypto is done before other muxing. Resolves #239 .
* Outgoing and Incoming connections are now state machines. Rather than being internally, automatically driven, they are driven by dialer.js and handler.js respectively. This allows finer control over transitions and helps provide clarity around when and where errors occur.
* Adds in some error standardization via the `err-codes` module
* Fixed readme links

## TODOs
- [x] There is redundant logic with the new Connection FSMs that needs to be cleaned up.
- [x] Add more, thorough tests for each of the connection FSMs.
- [x] Ensure incoming connections are being properly reused
- [x] General TODO cleanup
- [x] Add `class-is` module and remove instanceof logic, as it's unreliable across package versions.

## Thoughts on Future Additions
* As error codes are added throughout the system, have the FSMs handle fatal and non fatal errors appropriately.
* Add hooks for new connections to allow modules like the connection manager, to make intelligent decisions about those connections (dial backoffs, etc).

## References
* [Libp2p State Machine#98](https://github.com/libp2p/js-libp2p/issues/98)

Required by https://github.com/libp2p/js-libp2p/pull/257


jacobheun:
@diasdavid @pgte @vasco-santos I've still got some cleanup to do per my notes in the PR overview, but I'd love to get some feedback on the overall implementation of the FSM. There may be some considerations I've missed that we could roll into this update.

src/connection/index.js
```diff
@@ -0,0 +1,456 @@
...
+          this.emit('error', new Error(
+            `Circuit not enabled and all transports failed to dial peer ${this.theirB58Id}!`
+          ))
+          return this._state('disconnect')
+        }
+
+        if (circuitTried) {
+          this.emit('error', new Error(`No available transports to dial peer ${this.theirB58Id}!`))
+          return this._state('disconnect')
+        }
```
pgte, line 225: These last two conditions can dictate the end of the dialing attempt. As a client of the switch API, I'd like to now the reason why a dial failed. "No available transport" or "Circuit not enabled" doesn't tell me much of why the other transports failed.

As a remedy, I suggest that we keep track of the failing reasons and make them somehow available to the switch user.

pgte: As an alternative, emit a warning-like event like "connection attempt failed", containing an error object as payload.

src/connection/index.js
```diff
@@ -0,0 +1,456 @@
...
+
+          this.muxer = this.switch.muxers[key].dialer(conn)
+          this.switch.muxedConns[this.theirB58Id] = this
+
+          this.muxer.once('close', () => {
+            this._state('disconnect')
+          })
+
+          // For incoming streams, in case identify is on
+          this.muxer.on('stream', (conn) => {
```
pgte, line 377: can this `stream` event occurs multiple times?

src/dialer.js
```diff
@@ -0,0 +1,107 @@
...
+      connection = new ConnectionFSM({
+        _switch,
+        peerInfo,
+        muxer: _switch.muxedConns[b58Id] || null
+      })
+      connection.once('error', (err) => callback(err))
+      connection.on('connected', () => connection.protect())
+      connection.on('private', () => connection.encrypt())
+      connection.on('encrypted', () => connection.upgrade())
+      connection.on('muxed', () => {
```
pgte: Can all these events occur more than once?

pgte (commented):
LGTM, with some questions / comments.

src/dialer.js
jacobheun: No, they should only occur once, I'll update these.

src/connection/index.js
jacobheun: Yes. Once we have a muxed connection established it can open up any number of streams with that peer.

src/connection/index.js
jacobheun: I like the warning option as that would provide the user with more information about the various attempts, even if connections succeed on a later transport.

With the current structure the switch would need to emit the events. Perhaps the payload should also include the peerInfo so users understand where it came from. Something like:
```js
_switch.emit('error:connection_attempt_failed', {
  peerInfo: theirPeerInfo,
  error: ...
})
```

src/dialer.js
```diff
@@ -0,0 +1,107 @@
...
+      callback = protocol
+      protocol = null
+    }
+
+    callback = once(callback || function noop () {})
+
+    const peerInfo = getPeerInfo(peer, _switch._peerBook)
+    const b58Id = peerInfo.id.toB58String()
+
+    log(`${_switch._peerInfo.id.toB58String().slice(0, 8)} dial request to ${b58Id.slice(0, 8)} with protocol ${protocol}`)
```
vasco-santos: Verify if `protocol` exists and if not add `unknown` instead of allowing undefined

src/index.js
```diff
@@ -70,6 +79,43 @@ class Switch extends EE {
...
+    })
+    this.state.on('STARTED', () => {
+      log('The switch has started')
+      this.emit('started')
+    })
+    this.state.on('STOPPED', () => {
+      log('The switch has stopped')
+      this.emit('stopped')
+    })
+    this.state.on('error', (err) => this.emit('error', err))
```
vasco-santos: Aiming to ease debuggability, I think we should log the error here as well.

src/index.js
```diff
@@ -167,7 +167,7 @@ class Switch extends EE {
 
   /**
    * If a muxed Connection exists for the given peer, it will be closed
-   * and its reference on the Switch will be removed.
+   * and its reference on the Switch and will be removed.
```
vasco-santos: typo? I think it was correct before 🤔

src/index.js
```diff
@@ -197,6 +197,89 @@ class Switch extends EE {
...
+   * @fires Switch#error
+   * @returns {void}
+   */
+  _onStarting () {
+    each(this.availableTransports(this._peerInfo), (ts, cb) => {
+      // Listen on the given transport
+      this.transport.listen(ts, {}, null, cb)
+    }, (err) => {
+      if (err) {
+        return this.emit('error', err)
```
vasco-santos, line 239: log the error?

src/index.js
```diff
@@ -197,6 +197,89 @@ class Switch extends EE {
...
+    series([
+      (cb) => each(this.muxedConns, (conn, cb) => {
+        // If the connection was destroyed while we are hanging up, continue
+        if (!conn) {
+          return cb()
+        }
+
+        conn.muxer.end((err) => {
+          // If OK things are fine, and someone just shut down
+          if (err && err.message !== 'Fatal error: OK') {
```
vasco-santos: log the error?

src/limit-dialer/queue.js
```diff
@@ -37,15 +37,15 @@ class DialQueue {
    * @private
    */
   _doWork (transport, addr, token, callback) {
-    log('work')
+    log(`${transport.constructor.name}:work:start`)
     this._dialWithTimeout(transport, addr, (err, conn) => {
       if (err) {
-        log('work:error')
+        log(`${transport.constructor.name}:work:error`, err)
```
vasco-santos: use `log.error` here

vasco-santos (changes requested):
Seems good to me! Left just a few minor comments

src/connection/index.js
pgte: I see. This dialing is triggered by a `switch.dial()`, which is callback-based..
Having the switch emit these events independently of the dial() call means that, for the API user,  it will be hard to correlate a call with these warnings.
Usually I solve this by either returning an even stream or an event emitter. In this case, I think that something like a dialFSM would be something useful to consider...

src/dialer.js
pgte: 👍

src/connection/index.js
jacobheun: That's a good point. Currently `Libp2p` doesn't return anything from its dial methods so returning an event emitter there from a `Switch` dialFSM would enable us to keep backwards compatibility more easily while we transition the system.

For js-ipfs we'd also need to make sure to return that emitter from [`swarm.connect()`](https://github.com/ipfs/js-ipfs/blob/c73bd2f68a57651e2b7a67b522f79556040963ae/src/core/components/swarm.js#L71).

jacobheun: @pgte I've added a `dialsFSM` method to the switch, that can be used instead of `dial`. Instead of a connection, it will return the fsm. I've added some tests for it to show the events and functionality it has https://github.com/libp2p/js-libp2p-switch/pull/278/files#diff-080e21482eea9ffdd7ee511eb71d3522.

**The main things to note**:
The raw connection is still returned in the callback passed to dialFSM
1. Users can listen on `error` events to get fatal messages
2. Users can listen on `error:connection_attempt_failed` events to get a list of errors for failed transport dials, these are not fatal to the connection.
3. Users can listen on `close` events to know when the specific connection is closed.
4. Users can call `fsm.close()` to close that specific connection.

Are there other events or methods that would be helpful to have? We can always wait to add other things if there aren't higher priority nice to haves.

pgte: @jacobheun That's great! Other than successful connection attempt (which you can infer from having the "connected" callback called), I don't see any other so far.

jacobheun:
After reviewing the crypto logic in switch as it currently is, we're not actually supporting multiple crypto modules on dial, only on listen. This needs to be corrected, but after doing some small changes in this branch I think it should wait until this is done as it's a significant change. That will enable us to review those changes more easily to ensure they're done properly.

Also, since secio is currently our only crypto library in libp2p this isn't a critical need.

README.md
```diff
@@ -26,16 +26,17 @@ libp2p-switch is used by [libp2p](https://github.com/libp2p/js-libp2p) but it ca
...
 - [API](#api)
-  - [`switch.dial(peer, protocol, callback)`](#swarmdialpi-protocol-callback)
-  - [`switch.hangUp(peer, callback)`](#swarmhanguppi-callback)
-  - [`switch.handle(protocol, handler)`](#swarmhandleprotocol-handler)
-  - [`switch.unhandle(protocol)`](#swarmunhandleprotocol)
-  - [`switch.start(callback)`](#swarmlistencallback)
-  - [`switch.stop(callback)`](#swarmclosecallback)
-  - [`switch.connection`](#connection)
+  - [`switch.dial(peer, protocol, callback)`](#switchdialpeer-protocol-callback)
+  - [`switch.dialFSM(peer, protocol, callback)`](#switchdialfsmpeer-protocol-callback)
```
daviddias, line 31: better just mimic js-libp2p and do dial and dialProtocol

jacobheun: @diasdavid are you suggesting to add `dialProtocol` in addition to the new `dialFSM` or as a naming replacement? I think having `dial` handle calls with and without a protocol is reasonable overloading. Replacing it wouldn't be good as it has a different callback footprint to give users access to the ConnectionFSM for finer control/handling.  

I've exposed the same `dialFSM` method in the [libp2p branch](https://github.com/libp2p/js-libp2p/tree/feat/state-machine#libp2pdialfsmpeer-protocol-callback)

daviddias: We don't need to block this PR for this. What I mean is that given that libp2p (the module) is just an extension to the libp2p-switch (the dialing machine) that adds it other features, it would be cool if they both expose the same dialing/hangup APIs.

jacobheun: That makes sense. Ideally I'd like to deprecate `dialProtocol` in libp2p and just use `dial`. Having both feel unnecessary. I'll propose the change in an issue there and if `dialProtocol` ends up staying we can add that in here.

jacobheun:
@vasco-santos this should be ready for final review. I have this branch integrated in https://github.com/libp2p/js-libp2p/pull/257, which should also be ready for review.

README.md
```diff
@@ -26,16 +26,17 @@ libp2p-switch is used by [libp2p](https://github.com/libp2p/js-libp2p) but it ca
...
+  - [`switch.hangUp(peer, callback)`](#switchhanguppeer-callback)
+  - [`switch.handle(protocol, handler)`](#switchhandleprotocol-handler)
+  - [`switch.unhandle(protocol)`](#switchunhandleprotocol)
+  - [`switch.start(callback)`](#switchstartcallback)
+  - [`switch.stop(callback)`](#switchstopcallback)
+  - [`switch.connection`](#switchconnection)
```
vasco-santos: Can we please alphabetize the API? I believe this helps when looking for documentation.

vasco-santos: Can we please alphabetize the API? I believe this helps when looking for documentation.

src/index.js
```diff
@@ -197,6 +196,85 @@ class Switch extends EE {
     const transports = Object.keys(this.transports).filter((t) => t !== 'Circuit')
     return transports && transports.length > 0
   }
+
+  /**
+   * Issues a start on the Switch state.
+   *
+   * @param {function} callback deprecated: Listening for the `error` and `start` events are recommended
+   * @fires Switch#started
```
vasco-santos: `@fires Switch#started` ?

src/index.js
```diff
@@ -197,6 +196,85 @@ class Switch extends EE {
...
+      }
+      this.state('done')
+    })
+  }
+
+  /**
+   * A listener that will turn off all running services and listeners
+   *
+   * @private
+   * @fires Switch#error
```
vasco-santos: same as above

vasco-santos (commented):
@jacobheun all in all looks good to me 🙂 

Left just minor notes

jacobheun:
@vasco-santos I updated the order in the readme and fixed the improper usage of jsdocs `@fires` on individual methods and updated those events on the Switch class itself.

vasco-santos (approved):

PR #279, opened 2018-10-02
chore: update mplex and stats test numbers

jacobheun:
libp2p-mplex was updated with a fix for an [edge case issue](https://github.com/libp2p/js-libp2p-mplex/pull/84) where mplex's internal _send method was potentially doing two pushes (1. the header, 2. the length prefixed data) without respecting the `readableStream.push()`'s returned value.

Those potential two pushes are now consolidated into 1 push. This reduces the total number of messages that are sent from one peer to another. As each message being sent needs to be prefixed with [some additional information](https://github.com/libp2p/mplex#message-format), reducing the number of messages sent also reduces the total byte count.

This PR updates the stats tests total byte count to match the change created by consolidating `mplex._send()` into a single push.

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "aegir": "^15.1.0",
     "chai": "^4.1.2",
     "dirty-chai": "^2.0.1",
-    "libp2p-mplex": "~0.8.0",
+    "libp2p-mplex": "~0.8.2",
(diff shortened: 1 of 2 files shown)


vasco-santos (approved):
Nice @jacobheun ! 👍

PR #283, opened 2018-10-23
fix: ignore dial request when one is in progress

jacobheun:
In the instance that multiple dials are requested for the same connection, this will ignore subsequent dials. In the future, we should revisit this logic for coalescing dials (a successful request with dials in the queue) and backoffs (a failure with dials in the queue). There is a limit dialer at the transport level, but we don't currently handle it at the connection level.

This issue was exposed when testing the libp2p 0.24.0-rc.1 candidate release against the js-ipfs tests.


vasco-santos (approved):
LGTM!

PR #284, opened 2018-11-12
fix: dont dial an address that we have

jacobheun:
This PR filters out dialable addresses that match our addresses. This will help prevent us from dialing ourselves. 

This currently only handles strict equality. As we improve listen versus announce addresses, this should be improved to reflect that.

relates to https://github.com/libp2p/js-libp2p/issues/274

### DIFF
--- a/src/transport.js
+++ b/src/transport.js
@@ -97,7 +97,7 @@ class TransportManager {
     }
 
     // filter the multiaddrs that are actually valid for this transport
-    multiaddrs = TransportManager.dialables(transport, multiaddrs)
+    multiaddrs = TransportManager.dialables(transport, multiaddrs, this.switch._peerInfo)
     log('dialing %s', key, multiaddrs.map((m) => m.toString()))
 
     // dial each of the multiaddrs with the given transport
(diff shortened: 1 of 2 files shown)


vasco-santos (dismissed):
LGTM

vasco-santos:
@jacobheun seems like there are some problems with the tests

jacobheun:
Yep, looking into it.

blakebyrnes:
Is it worth using ipfs format for the test?  I believe the addresses related to the issue logged on js-libp2p project will have duplicate ip and port, but different peerId (https://github.com/libp2p/js-libp2p/issues/274)

jacobheun:
@vasco-santos looks like the latest secio is causing some odd behaviors on linux, I'm working on getting those resolved.

@blakebyrnes great point, I'll add some more verbose testing of addresses to this once I resolve the other issue.

jacobheun:
I've updated the logic here and expanded the addresses a bit. Doing a comparison of the host and port is problematic as they're not easy to get from the addresses without more complex logic. This now attempts to use multiaddr's decapsulate functionality to find our address in the address to be dialed. If it's there, we avoid dialing. This isn't a perfect solution but should help mitigate some of the self dials until we add https://github.com/libp2p/js-libp2p/issues/202 and can do better checking.

This needs #285 to resolve the issue with the tests.

jacobheun:
@vasco-santos ci is green, this is ready for a re-review. Thanks!

vasco-santos (approved):
LGTM

PR #285, opened 2018-11-14
Improve connection error emitting and verify identify connection close on failure

jacobheun:
This improves error handling for the latest state machine update and fixes test issues in master.


jacobheun:
@vasco-santos this will resolve the testing problem from #284

vasco-santos (approved):
Looks good!

We are a small step away from the coverage, but I think we do not need to get it now, in order to not block the `libp2p` release longer

PR #286, opened 2018-11-15
fix: ensure circuit listens last on start

jacobheun:
Currently Switch starts all available transports listening on startup in parallel. The problem with this is that circuit should listen last. This gives the transports the opportunity to update addresses. For example, in js-ipfs, one of the default addresses in node is `/ip4/0.0.0.0/tcp/4002`.  The tcp transport will see `0.0.0.0` and bind it to available ip addresses on the network and replace it to something like `/ip4/192.168.X.X/tcp/4002`. 

Currently, the circuit address will end up looking like `/p2p-circuit/ip4/0.0.0.0/tcp/4002` instead of `/p2p-circuit/ip4/192.168.X.X/tcp/4002`, this fixes that.

I've updated a test in the circuit relay suite to verify the replacement so we avoid this in the future.

Requires #285 to resolve ci failures.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -3,6 +3,7 @@
 const FSM = require('fsm-event')
 const EventEmitter = require('events').EventEmitter
 const each = require('async/each')
+const eachSeries = require('async/eachSeries')
 const series = require('async/series')
 const TransportManager = require('./transport')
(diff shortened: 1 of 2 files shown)


vasco-santos (approved):
👍

PR #289, opened 2018-12-13
Use retimer to avoid creating so many timers

mcollina:
This PR does three things

* Use retimer to reduce the number of timers that needs to be created.
* Remove an unneeded `setImmediate`
* Fixes linting


jacobheun (approved):
This looks good, I'll fix the commitlint on squash

mcollina:
Wow this was quick!

Issue #288, opened 2018-12-11
switch doesn't always clean up connections


jacobheun:
Duplicate of https://github.com/libp2p/js-libp2p/issues/295

We need to appropriately handle multiple connections to a single peer. Currently switch is only tracking a 1:1 mapping of `peerid:connection`.

jacobheun:
Resolved in the latest release 0.41.3

PR #291, opened 2018-12-13
improve connection close management

jacobheun:
**Handling multiple connections to a peer**
Switch was previously only tracking 1 connection per peer, however it's possible to have more than 1 connection per peer.  This adds the ability to track multiple connections per peer, but will attempt to use a single connection when possible.

**Closing base connections** #288
Switch was previously not properly closing the base connection. This could result in streams staying open on the receiver side if they allow half open connections. Now when a connection is closing, we abort the base pull stream connection to notify the receiver that we're done.

* Ensure there is a transport before filtering dialables.
* Stat starting and stopping are better tied into the switch start and stop


vasco-santos (approved):
:+1:

PR #295, opened 2018-12-19
refactor stat to not use forEach

mcollina:
forEach is 10x slower than a regular for(;;) loop, and it should
be avoided in hot code paths.

### DIFF
--- a/src/stats/stat.js
+++ b/src/stats/stat.js
@@ -25,14 +25,18 @@ class Stats extends EventEmitter {
 
     this._update = this._update.bind(this)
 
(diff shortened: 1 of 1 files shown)


jacobheun:
@mcollina any particular reason var is being used in the for loops instead of let and const? Other than that this looks good.

mcollina:
Historically var has been faster than let/const. In latest V8 versions, this has been optimized. I’m being conservative especially because of browser environments.

jacobheun:
Ah, that is good to know, thanks!

jacobheun (approved):

Issue #287, opened 2018-12-10
Maximum call stack size exceeded


alanshaw:
```js
/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/observer.js:41
    peerInfo.then((_peerInfo) => {
             ^

RangeError: Maximum call stack size exceeded
    at Promise.then (<anonymous>)
    at willObserve (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/observer.js:41:14)
    at pull.map (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/observer.js:34:9)
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/pull-stream/throughs/map.js:13:23
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/pull-reader/index.js:114:13
    at /Users/alan/Code/protocol-labs/js-ipfs/node_modules/pull-reader/index.js:114:13
    at onNext (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/pull-catch/index.js:17:17)
    at drain (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/stream-to-pull-stream/index.js:141:18)
    at Channel.<anonymous> (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/stream-to-pull-stream/index.js:150:5)
    at Channel.emit (events.js:182:13)
```

jacobheun:
@alanshaw any other info on the state of the node when this happened? Testing/idle node/etc.

alanshaw:
Sorry, should lave left more context - just wanted to get this logged before I forgot!

It's a JS IPFS daemon running on this branch https://github.com/ipfs/js-ipfs/pull/856

I'm using companion and I opened the WebUI, looked around a bit and clicked on the peers page to see the hundreds of nodes (very exciting btw), shortly after my daemon crashed with that error.

alanshaw:
Biggest stack I have ever seen https://ci.ipfs.team/blue/rest/organizations/jenkins/pipelines/IPFS/pipelines/js-ipfs/branches/PR-856/runs/80/nodes/257/steps/386/log/?start=0

achingbrain:
Seeing this in Jenkins again today:

```
[...]

PR #297, opened 2018-12-20
clean up sync in async usage

jacobheun:
This cleans up some logic in the identify flow where sync calls were being done in async code, which can lead to stack bloating. This moves the sync logic out of the async control flow.

Relates to https://github.com/libp2p/js-libp2p-switch/issues/287


src/connection/manager.js
```diff
@@ -145,20 +145,18 @@ class ConnectionManager {
...
+            ], (err, peerInfo, observedAddrs) => {
               if (err) {
                 return muxedConn.end(() => {
                   callback(err, null)
                 })
               }
 
+              observedAddrs.forEach((oa) => {
+                this.switch._peerInfo.multiaddrs.addSafe(oa)
+              })
```
mcollina, line 152: I would recommend using a for loop here instead

src/connection/manager.js
```diff
@@ -145,20 +145,18 @@ class ConnectionManager {
...
-              (conn, cb) => identify.dialer(conn, cryptoPI, cb),
-              (peerInfo, observedAddrs, cb) => {
-                observedAddrs.forEach((oa) => {
-                  this.switch._peerInfo.multiaddrs.addSafe(oa)
-                })
-                cb(null, peerInfo)
-              }
-            ], (err, peerInfo) => {
+              (conn, cb) => identify.dialer(conn, cryptoPI, cb)
+            ], (err, peerInfo, observedAddrs) => {
```
mcollina: You might want to remove that waterfall call, it would improve things as well.

jacobheun:
I refactored the waterfall logic to use awaits and created some utility methods to make that cleaner. Also removed `forEach` in favor of `for(;;)`

vasco-santos (approved):
LGTM!

dirkmc:
With regards to switching forEach loops to for loops etc, I would recommend instead focusing performance improvements on changes that make a difference in terms of big O notation: https://en.wikipedia.org/wiki/Big_O_notation

src/connection/manager.js
```diff
@@ -136,34 +136,33 @@ class ConnectionManager {
...
+              results = await identifyDialer(msConn, cryptoPI)
+            } catch (err) {
+              return muxedConn.end(() => {
+                callback(err, null)
+              })
+            }
+
+            const { peerInfo, observedAddrs } = results
+
+            for (var i = 0; i < observedAddrs.length; i++) {
```
dirkmc, line 157: Can you use a `for ... of` loop here instead?

src/utils.js
```diff
@@ -0,0 +1,49 @@
+'use strict'
```
dirkmc, line 1: Can you use promisify instead of these functions? https://github.com/digitaldesignlabs/es6-promisify#readme

jacobheun: I added these as an interim solution. We're going to be doing a lot of work soon for they async iterators initiative https://github.com/libp2p/js-libp2p/issues/266 which will remove the need for these or promisfy. If that ends up taking longer than we'd like, or we have more immediate needs in other parts of the code base, then I think switching over to promisify is warranted.

dirkmc: 👍

src/connection/manager.js
mcollina: `for ... of ` suffers from some performance issues still.

jacobheun:
In regards to overall performance improvements, we have an OKR for Q1 to get benchmarking in place for libp2p. That along with the current work for ipfs benchmarking will put us in a better position to do isolated analyses to see where the bottlenecks are in switch and the rest of libp2p. While I think those are very important, I want to avoid pushing more of that into this particular PR, as it's goal was to fix an issue with synchronous code being executed asynchronously, to help alleviate #287. I appreciate all the great feedback!

mcollina:
> With regards to switching forEach loops to for loops etc, I would recommend instead focusing performance improvements on changes that make a difference in terms of big O notation: https://en.wikipedia.org/wiki/Big_O_notation

I agree, the main gains are in algorithmic improvements in big O notation. However, there is a lot to gain from making the code more optimizable by V8.

As an example, `forEach` would do  `n` function calls, which would not appear in a big O analysis. These have a cost in term of memory allocation and work inside V8, that would not bubble up when just reasoning about algorithms. Each one of these can easily eat 5-10% of the time of a benchmark in a hot path.

This is not always a hot path, but I've see this function in a significant number of flamegraphs.

PR #298, opened 2018-12-21
Add connection type param to hangUp()

### LINKED ISSUE #299: Option to specify "outbound connections only" in libp2p.hangUp()
In peer-base each peer keeps connections open to the peers in its [Dias set](https://github.com/peer-base/peer-base/blob/master/docs/PROTOCOL.md#computing-the-set-of-target-peers). The Dias set membership changes as peers come online and go offline.

For example:
- Peer A has a Dias set that includes Peer B
- Membership changes such that Peer A's Dias set no longer includes Peer B
- Peer A should disconnect its outbound connection to peer B
- However if Peer B's Dias set includes Peer A, then that connection should be maintained

So from Peer A's perspective, it only wants to disconnect the outbound connection, leaving any inbound connection intact.

Later, when Peer A shuts down, it wants to disconnect both inbound and outbound connections.

Note that the `peer:connect` event doesn't indicate if a connection is inbound or outbound, so it's not possible for peer base to keep track of inbound vs outbound connections, and in any case it would simplify the code a lot if we could rely on libp2p to do that instead.

dirkmc:
Fixes https://github.com/libp2p/js-libp2p/issues/299

Note: Requires https://github.com/libp2p/js-libp2p-circuit/pull/41 for tests to pass

### DIFF
--- a/test/dial-fsm.node.js
+++ b/test/dial-fsm.node.js
@@ -196,7 +196,12 @@ describe('dialFSM', () => {
       expect(meta.type).to.eql(Switch.ConnectionType.Incoming)
     })
 
-    const conn = switchA.dialFSM(switchB._peerInfo, '/parallel/1.0.0', () => {
+    // Dial from A -> B and wait for connections to be established in both
+    // directions
+    awaitEvents([
+      [switchA, 'peer-mux-established', 2],
+      [switchB, 'peer-mux-established', 2]
+    ], () => {
       // Both should have 1 incoming / 1 outgoing connection
       expect(switchA.connection.getAll()).to.have.length(2)
       expect(switchB.connection.getAll()).to.have.length(2)
@@ -206,6 +211,7 @@ describe('dialFSM', () => {
         expect(err).to.not.exist().mark()
       })
     })
+    const conn = switchA.dialFSM(switchB._peerInfo, '/parallel/1.0.0')
 
     // Hold the dial from A, until switch B is done dialing to ensure
     // we have both incoming and outgoing connections
@@ -243,7 +249,12 @@ describe('dialFSM', () => {
       expect(meta.type).to.eql(Switch.ConnectionType.Outgoing)
     })
 
-    const conn = switchA.dialFSM(switchB._peerInfo, '/parallel/1.0.0', () => {
+    // Dial from A -> B and wait for connections to be established in both
+    // directions
+    awaitEvents([
+      [switchA, 'peer-mux-established', 2],
+      [switchB, 'peer-mux-established', 2]
+    ], () => {
       // Both should have 1 incoming / 1 outgoing connection
       expect(switchA.connection.getAll()).to.have.length(2)
       expect(switchB.connection.getAll()).to.have.length(2)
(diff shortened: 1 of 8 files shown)


jacobheun:
@dirkmc I reran CI so the latest `libp2p-circuit` would get pulled in. Looks like there are some failures, can you take a look at those?

dirkmc:
Thanks Jacob, I’m in Nicaragua this week but I can take a look when I’m
back next Wednesday

dirkmc:
@jacobheun I made the tests more consistent so it looks like everything is passing now

test/dial-fsm.node.js
```diff
@@ -203,3 +315,19 @@ describe('dialFSM', () => {
     })
   })
 })
+
+function awaitEvents (defs, cb) {
```
jacobheun: One request, this is a useful function that should probably get used in some of the other test suites (not in this PR). Can we move this to a new utils index file in the test folder?  

Other than that, this looks great!

dirkmc: I saw there is already a test/utils.js file so I put it in there, let me know if that works

jacobheun:
I ran this against the libp2p test suites and encountered a couple test failures with the circuit relay tests there. I am taking a look at those to see what's going on.

dirkmc:
@jacobheun did you have any luck finding why the libp2p tests are failing?

jacobheun:
I haven't had a chance to look at in too much depth.  I'm a bit unavailable this week, but will be able to look at it next week if I can't get to it later this week.

dirkmc:
Great, thank you

jacobheun:
Okay, I had some time today to look into the issue. The failure with the circuit relay tests in the js-libp2p repo is due to the change in dialer:
```js
let connection = _switch.connection.getOne(b58Id) // old
let connection = _switch.connection.getOne(b58Id, ConnectionFSM.ConnectionType.Outgoing) // new
```

Two nodes dial to the relay, both will have outgoing connections. When nodeA attempts to dial to nodeB over the relay the connection from nodeA -> Relay is fine. When the Relay attempts to connect to nodeB, it has an issue because it doesn't use the existing connection to nodeB, since that connection is `Incoming` for the relay. If the relay is passive, it won't attempt dialing so the request fails. An active relay should dial, but it looks like there is a bug there that I will create a separate issue for.

Ideally, I think it would be better to not have the concept of incoming vs outgoing and instead use a goodbye protocol to manage scenarios like that mentioned in https://github.com/libp2p/js-libp2p/issues/299#issue-392283726. As a workaround, we had previously discussed using `both` as a type. Perhaps an alternative would be to remove type and start using a tags attribute. Tagging is eventually going to be needed by Connection Manager, so this could be the starting point for that. When you go to hang up "outgoing" connections you could check the tags. If it also has incoming, just remove the outgoing tag, otherwise hangup.

I need to think about this a bit more.

dirkmc:
hmm yes it's more complicated that I had imagined :)
Let me know once you've thought it through what makes most sense in the wider context of libp2p, it seems like there may be some bigger structural changes to be made. I'm happy to help out with writing tests and implementation.

daviddias:
@jacobheun with the refactor, these need to be remade or closed. I'll let you do the review

jacobheun:
Closing this as tagging will be needed for this and it will likely change the structure of this significantly.

PR #302, opened 2019-01-09
TransportManager Refactoring to fix #300

### LINKED ISSUE #300: Refactor TransportManager to avoid storing listeners directly on the Transport object
TransportManager [stores listeners](https://github.com/libp2p/js-libp2p-switch/blob/master/src/transport.js#L44) on the underlying Transport object itself. This causes strange things to happen if, for example, the Transport implementation extends EventEmitter ([as it does in peer-base](https://github.com/peer-base/peer-base/blob/master/src/transport/app-transport.js#L13)) because EventEmitter has a [method called `listeners()`](https://nodejs.org/api/events.html#events_emitter_listeners_eventname) which gets clobbered.

Currently Switch has an [object called `transports`](https://github.com/libp2p/js-libp2p-switch/blob/master/src/index.js#L41) whose keys are transport names (eg 'tcp') and values are the Transport implementations themselves. I suggest adding a level of indirection such that the values have an object like this:
```Javascript
{
  transport: <Transport implementation>,
  listeners: [<Listener>, <Listener>, ...]
}
```
The only complicating factor would be if there are libraries that depend on libp2p-switch and reference `Switch.transports`. As far as I can tell there are no references elsewhere in libp2p.

nikor:
Encapsulates transports in TransportManager and moves the listeners outside the transport object.

One of the test in js-libp2p fails with this pull request. There is a fix for that test on this branch: https://github.com/nikor/js-libp2p/tree/js-libp2p-switch_300

### DIFF
--- a/.aegir.js
+++ b/.aegir.js
@@ -38,8 +38,8 @@ function pre (done) {
       peerA.multiaddrs.add(maA)
       switchA = new Switch(peerA, new PeerBook())
 
-      switchA.transport.add('ws', new WebSockets())
-      switchA.transport.listen('ws', {}, echo, cb)
+      switchA.transportManager.add('ws', new WebSockets())
+      switchA.transportManager.listen('ws', {}, echo, cb)
     })
   }
 
@@ -53,7 +53,7 @@ function pre (done) {
       peerB.multiaddrs.add(maB)
       switchB = new Switch(peerB, new PeerBook())
 
-      switchB.transport.add('ws', new WebSockets())
+      switchB.transportManager.add('ws', new WebSockets())
       switchB.connection.addStreamMuxer(spdy)
       switchB.connection.reuse()
       switchB.handle('/echo/1.0.0', echo)
@@ -72,7 +72,7 @@ function pre (done) {
 
 function post (done) {
   parallel([
-    (cb) => switchA.transport.close('ws', cb),
+    (cb) => switchA.transportManager.close('ws', cb),
     (cb) => switchB.stop(cb),
     (cb) => sigS.stop(cb)
   ], done)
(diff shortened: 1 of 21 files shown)


dirkmc:
I would suggest using a `Map` for `TransportManager.transports`. This will allow you to simplify some of the other code. For example `getTransports()` can be implemented as
```Javascript
return new Map([...this.transports].map(([k, v]) => [k, v.transport]))
```

nikor:
I will change it to a map, and i will try to make test pass as well :)

jacobheun:
@nikor tests seem to be okay. The commitlint is the main thing that's failing. (Code coverage is also "failing", but it's a nominal drop in percentage).  When committing dont capitalize the first letter of the commit message. `fix: Remove listeners from transport object` should be `fix: remove listeners from transport object`. It's a bit silly, but it keeps the changelog consistent.

Since this is going to be a breaking change, I think we should take advantage of the change to naming transport related things. Right now we have things like `switch.transport.transports[key].transport`, which makes it pretty easy to get lost. Some quick thoughts on possible changes:

Use `switch.transportManager` instead of `switch.transport` for easier reading.
`switch.transportManager.get(key)` and maybe privatize the `transportManager._transports` map
`switch.transportManager.getAll()` instead of `getTransports`

Thanks for submitting a PR for this!

nikor:
@jacobheun Thank you for explaining why commitlint failed. I will stick to lowercase in the future :)
I will make a new commit with the naming changes.

nikor:
Okay, i think the api changes is done.
I have make the changes need to js-libp2p on this branch https://github.com/nikor/js-libp2p/tree/js-libp2p-switch_300

daviddias:
@jacobheun with the refactor, these need to be remade or closed. I'll let you do the review

PR #303, opened 2019-01-26
Import async/setImmediate to avoid webpack errors

PiusNyakoojo:


jacobheun (approved):
Looks good, thanks for submitting the fix!

PR #305, opened 2019-02-14
test: add pull-mplex to test suite

jacobheun:
* Adds pull-mplex to some of the tests
* Changes CI over to travis

**Note** The code coverage complaint is about the .aegir.js file changes, which can be ignored.


test/swarm-muxing.node.js
```diff
@@ -9,7 +9,7 @@ chai.use(dirtyChai)
 const parallel = require('async/parallel')
 const TCP = require('libp2p-tcp')
 const WebSockets = require('libp2p-websockets')
-const spdy = require('libp2p-spdy')
+const spdy = require('pull-mplex')
```
vasco-santos: why changing `spdy` to `pull-mplex` here?

vasco-santos (changes requested):

test/swarm-muxing.node.js
jacobheun: Good catch! I was testing locally and forgot to fix that. Now the tests will run against both mplex and spdy

vasco-santos (approved):
LGTM!

PR #306, opened 2019-03-05
fix: dial in series until we have proper abort support

jacobheun:
Currently the switch will dial all addresses for a particular transport in parallel. Since we don't have any real abort support for when a dial succeeds, we should resort to dialing in serial to avoid flooding a peer with unused connections. This may result in slower dials to peers, but should also prevent flooding connections. As we improve the logic of dialing, and add proper abort support, we should be able to dial in parallel, select the best connection and abort the rest.

The first connection to succeed will be used. Circuit is already dialed last, so we don't need to worry about using Circuit if we're able to dial the node directly. 

----
This also improves the stability of the transport tests by moving `tryEcho` calls inside the dial callbacks, and adds an `after` handler to ensure the switch's are properly stopped after each transport suite is finished.

### DIFF
--- a/src/connection/index.js
+++ b/src/connection/index.js
@@ -197,7 +197,10 @@ class ConnectionFSM extends BaseConnection {
     const tKeys = this.switch.availableTransports(this.theirPeerInfo)
 
     const circuitEnabled = Boolean(this.switch.transports[Circuit.tag])
-    let circuitTried = false
+
+    if (circuitEnabled && !tKeys.includes(Circuit.tag)) {
+      tKeys.push(Circuit.tag)
+    }
 
     const nextTransport = (key) => {
       let transport = key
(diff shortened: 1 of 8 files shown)


src/connection/index.js
```diff
@@ -208,23 +211,20 @@ class ConnectionFSM extends BaseConnection {
           ))
         }
 
-        if (circuitTried) {
-          return this.close(Errors.CONNECTION_FAILED(
-            new Error(`No available transports to dial peer ${this.theirB58Id}!`)
-          ))
-        }
+        return this.close(Errors.CONNECTION_FAILED(
+          new Error(`No available transports to dial peer ${this.theirB58Id}!`)
```
vasco-santos: add `errcode` here?

vasco-santos (commented):
Awesome jacob! This looks good and will really help now!

Just added a minor comment here

src/connection/index.js
jacobheun: `Errors.CONNECTION_FAILED` is a wrapper function that generates `errCode` errors.

https://github.com/libp2p/js-libp2p-switch/blob/v0.41.5/src/errors.js

vasco-santos: Ok, my bad. So, we do not need to create the `new Error` here. It can be omitted and `errCode` will handle it.

vasco-santos (approved):

PR #307, opened 2019-03-05
feat: add protocols to PeerInfo after muxing on dial

### LINKED ISSUE #304: Storing protocols that peers support
Recently, I created [this issue in js-peer-info](https://github.com/libp2p/js-peer-info/issues/72) about storing protocols that peers support. @jacobheun suggested that js-libp2p-switch needs to be updated to perform `ls` on `multistreams` in order to set an initial protocol list for peers in such a way that this list can be updated later.

jacobheun:
This adds support on dial to run `ls` from `multistream-select`. This will enable things like pubsub and the dht to check the protocols a peer supports before attempting any actions on that peer.

Ideally the ls would occur much earlier in the dial, so that crypto and muxer selection could be decided based on the ls. However, due to https://github.com/multiformats/js-multistream-select/issues/47, this has been moved to after the muxer upgrade. Once that issue is fixed, the ls should occur at the start of the connection to allow for more efficient protocol selection.

Resolves #304


src/connection/index.js
```diff
@@ -381,16 +381,60 @@ class ConnectionFSM extends BaseConnection {
             this.switch.protocolMuxer(null)(conn)
           })
 
-          this.switch.emit('peer-mux-established', this.theirPeerInfo)
-
-          this._didUpgrade(null)
+          // TODO: this should really happen much earlier. However, currently
```
vasco-santos, line 384: Can you open an issue here for this change?

vasco-santos (approved):
Great 🎉 

LGTM!

Mikerah:
One thing I noticed is that you didn't add any extra tests to `connection.node.js` and `stream-muxers.node.js`. Why is that? Doesn't this change also affect `ConnectionFSM`?

jacobheun:
> One thing I noticed is that you didn't add any extra tests to connection.node.js and stream-muxers.node.js. Why is that? Doesn't this change also affect ConnectionFSM?

dialFSM leverages ConnectionFSM. We could add tests to both, but it's a bit redundant. We're detecting the `muxed` event from the ConnectionFSM, so we should be able to reasonably assume it's working properly.

`connection.node.js` and `stream-muxers.node.js` will also ultimately use the same code, and the dialFSM tests are leveraging a fully equipped Switch for dialing, so it shouldn't be needed.

The big thing I'd like to do before releasing this is to test it against a go-libp2p node to verify the interop is working properly, since we haven't used `ls` in js yet.

Mikerah:
Any updates on getting the interop tests done?

raulk:
Did I understand correctly that the dialing machine will own the dial until `ls` has run and we have discovered the protocols supported by our peer? This would result in 1-RTT extra until we yield the dial to the caller?

go-libp2p behaves differently. Identify runs asynchronously after connection upgrade. Protocols relying on "new connection" events (DHT, Pubsub, etc.) try to open a stream for their proto immediately, thus kicking off Multistream 1.0 protocol negotiation on the stream. If the other party NACKS, the protocol ignores that peer and closes the stream. 

With this PR we avoid kicking a protocol into action at the cost of 1-RTT. I'm not sure it's worth it, since streams are cheap and even if we have upfront knowledge that the protocol is supported, MS 1.0 negotiation has to occur anyway (MS 2.0 does away with this).

jacobheun:
@Mikerah not yet, sorry about the delay, I've been working on some performance issues for the upcoming jsipfs release.

My thought for interop was to link this branch in a new test at https://github.com/libp2p/interop. The new test being a simple `connect` for js -> go, followed by a check of the protocols on that PeerInfo object from the js daemon's libp2p node. We'll only be able to verify the JS node is able to get the protocols, since we don't have readily available access to the protocols go is storing for peers, but that should be sufficient for this.

I don't think I will be able to get to this until later tomorrow at the earliest. If you'd like to try adding that test in the interim that would be helpful.

jacobheun:
> Did I understand correctly that the dialing machine will own the dial until ls has run and we have discovered the protocols supported by our peer? This would result in 1-RTT extra until we yield the dial to the caller?

Yes. This probably should just happen asynchronously and potentially just get moved into js-libp2p-identify. 

The idea is that doing it at the beginning, once the ls issue is fixed in multistream-select, you avoid potential round trips needed for crypto / stream muxer negotiation. Right now, this probably isn't a problem for us, because that's fairly limited. js-ipfs for example is going to try to negotiation `/secio/1.0.0` which should succeed, followed by `/mplex/x.x.x` which should succeed. There are instances where people may have their stream muxer order different in which case `/spdy/x.x.x` might try to be negotiated first, which may not be supported. 

Since we're not really benefitting from the knowledge up front, we can make it asynchronous until either we move to multistream2, or we get additional crypto/muxers that warrant us having that knowledge up front.

> go-libp2p behaves differently.

It would be great if we had recommendations for implementations to accompany specs. With more languages implementing libp2p, using go as a reference implementation is likely going to scale poorly. It also has the risk of people replicating logic that perhaps we'd like to change/update, but haven't gotten to yet.

jacobheun:
Closing this in favor of #311, which includes an update to Identify to get the supported protocols. This is in line with how go functions.

PR #309, opened 2019-03-09
fix: peer disconnect event and improve logging performance

jacobheun:
The `peer-mux-closed` event was being emitted on connection close, even when no muxed connection was established. Since the connect event ,`peer-mux-established`, is emitted only after muxing, this PR makes the disconnect event in line with the connect event.

This PR also moves any formatting of debug logs into the responsibility of the debugger. The `debug` module immediately checks whether or not it is enabled for each logger namespace before doing any formatting, and returns early if not. This will prevent unneeded string formatting when debugging is not enabled, which will help improve performance.

### DIFF
--- a/src/connection/base.js
+++ b/src/connection/base.js
@@ -22,7 +22,7 @@ class BaseConnection extends EventEmitter {
    */
   close (err) {
     if (this._state._state === 'DISCONNECTING') return
-    this.log(`closing connection to ${this.theirB58Id}`)
+    this.log('losing connection to %s', this.theirB58Id)
     if (err && this._events.error) {
       this.emit('error', err)
     }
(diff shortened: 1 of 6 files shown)


vasco-santos (approved):
Great work @jacobheun! 

LGTM

src/connection/base.js
```diff
@@ -22,7 +22,7 @@ class BaseConnection extends EventEmitter {
    */
   close (err) {
     if (this._state._state === 'DISCONNECTING') return
-    this.log(`closing connection to ${this.theirB58Id}`)
+    this.log('losing connection to %s', this.theirB58Id)
```
dirkmc: Missing the `c` here :)

alanshaw (approved):
Code change looks sensible and tests all still pass!

PR #310, opened 2019-03-14
feat: add basic dial queue to avoid many connections to peer

jacobheun:
See the commit details, added here for ease:

> BREAKING CHANGE: This adds a very basic dial queue peer peer. This will prevent multiple, simultaneous dial requests to the same peer from creating multiple connections. The requests will be queued per peer, and will leverage the same connection when possible. The breaking change here is that `.dial`, will no longer return a connection. js-libp2p, circuit relay, and kad-dht, which use `.dial` were not using the returned connection. So while this is a breaking change it should not break the existing libp2p stack. If custom applications are leveraging the returned connection, they will need to convert to only using the connection returned via the callback.

Required by https://github.com/libp2p/js-libp2p/pull/336


src/dialer/index.js
```diff
@@ -0,0 +1,72 @@
...
+  const dialerQueue = new DialerQueue(_switch)
+
+  _switch.state.on('STOPPING:enter', abort)
+
+  /**
+   * @param {object} options
+   * @param {PeerInfo} options.peerInfo
+   * @param {string} options.protocol
+   * @param {boolean} options.useFSM If `callback` should return a ConnectionFSM
+   * @param {function(Error, Connection)} options.callback
```
vasco-santos: isn't it weird to have the callback in the `options` object?

src/dialer/queue.js
```diff
@@ -0,0 +1,244 @@
...
+   * and executes there callback with an error.
+   *
+   * This causes the entire DialerQueue to be drained
+   */
+  abort () {
+    const queues = Object.values(this._queue)
+    queues.forEach(queue => {
+      while (queue.size() > 0) {
+        let dial = queue.shift()
+        dial.callback(new Error('DIAL_ABORTED'))
```
vasco-santos: add `err-code` here

src/dialer/queue.js
```diff
@@ -0,0 +1,244 @@
...
+   *
+   * Returns an array containing two items. The ConnectionFSM and wether
+   * or not the ConnectionFSM was just created. The latter can be used
+   * to determine dialing needs.
+   *
+   * @private
+   * @param {PeerInfo} peerInfo
+   * @returns {[ConnectionFSM, Boolean]}
+   */
+  getOrCreateConnection (peerInfo) {
```
vasco-santos:
```suggestion
  _getOrCreateConnection (peerInfo) {
```

src/dialer/queue.js
```diff
@@ -0,0 +1,244 @@
...
+    return [connectionFSM, didCreate]
+  }
+
+  /**
+   * Executes the next dial in the queue for the given peer
+   * @private
+   * @param {PeerInfo} peerInfo
+   * @returns {void}
+   */
+  run (peerInfo) {
```
vasco-santos:
```suggestion
  _run (peerInfo) {
```

vasco-santos (commented):
Overall looks good! Awesome work Jacob!

src/dialer/index.js
jacobheun: haha, yes, you're correct. I think it makes sense for it to go into the actual queue like that. Options really isn't the right term, it's really a DialRequest.

src/connection/handler.js
```diff
@@ -14,31 +14,32 @@ function listener (_switch) {
...
 
-      log('received incoming connection')
-      const connFSM = new IncomingConnection({ connection, _switch, transportKey })
+        connFSM.once('error', (err) => log(err))
+        connFSM.once('private', (_conn) => {
+          // Use the custom handler, if it was provided
+          if (handler) {
+            return handler(_conn)
+          }
+          connFSM.encrypt()
```
alanshaw, line 37: The calls to `protect()` and `encrypt()` have swaped places, was that intentional? What effect does that have? Seems out of place for this PR?

src/connection/index.js
```diff
@@ -366,8 +379,6 @@ class ConnectionFSM extends BaseConnection {
           const conn = observeConnection(null, key, _conn, this.switch.observer)
 
           this.muxer = this.switch.muxers[key].dialer(conn)
-          // this.switch.muxedConns[this.theirB58Id] = this
-          this.switch.connection.add(this)
```
alanshaw, line 370: Why is this no longer added?

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+   * Adds the `dialRequest` to the queue and ensures the queue is running
+   *
+   * @param {DialRequest} dialRequest
+   */
+  add ({ peerInfo, protocol, useFSM, callback }) {
+    const id = peerInfo.id.toB58String()
+    const proxyConnection = new Connection()
+    proxyConnection.setPeerInfo(peerInfo)
+
+    callback = once(callback || noop)
```
alanshaw:
```suggestion
    callback = callback ? once(callback) : noop
```

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+
+  /**
+   * Iterates over all items in the DialerQueue
+   * and executes there callback with an error.
+   *
+   * This causes the entire DialerQueue to be drained
+   */
+  abort () {
+    const queues = Object.values(this._queue)
+    queues.forEach(queue => {
```
alanshaw: Each queue should have an `abort` method you can call, this should stop queue processing and call each of the outstanding dial callbacks.

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+    start: function () {
+      isRunning = true
+    },
+    stop: function () {
+      isRunning = false
+    }
+  }
+}
+
+class DialerQueue {
```
alanshaw: I think this name is a bit misleading, this is actually an interface to many per-peer dial queues.

As a suggestion, it might be a bit easier to reason about if we could use it like this:

```js
// Get or create the dial queue for this peer
const dialQueue = dialQueueManager.getQueue(peerInfo)

// Queue already knows it's peerInfo so no need to pass
dialQueue.add(protocol, useFSM, callback)

// Later
dialQueue.abort()
```

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+    return [connectionFSM, didCreate]
+  }
+
+  /**
+   * Executes the next dial in the queue for the given peer
+   * @private
+   * @param {PeerInfo} peerInfo
+   * @returns {void}
+   */
+  _run (peerInfo) {
```
alanshaw: Queue running logic should be encapsulated in each queue.

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+    const id = peerInfo.id.toB58String()
```
alanshaw: Cache this when the queue is created so you don't have to encode for every dial.

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+   * @returns {[ConnectionFSM, Boolean]}
+   */
+  _getOrCreateConnection (peerInfo) {
+    const id = peerInfo.id.toB58String()
+    let connectionFSM = this.switch.connection.getOne(id)
+    let didCreate = false
+
+    if (!connectionFSM) {
+      connectionFSM = new ConnectionFSM({
+        _switch: this.switch,
```
alanshaw, line 137: I don't understand the occasional `_` in parameter names. It's usually used to denote a "private" member on an object, but this doesn't make sense when passed as a parameter.

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+
+      // Add control events and start the dialer
+      connectionFSM.once('connected', () => connectionFSM.protect())
+      connectionFSM.once('private', () => connectionFSM.encrypt())
+      connectionFSM.once('encrypted', () => connectionFSM.upgrade())
+
+      didCreate = true
+    }
+
+    return [connectionFSM, didCreate]
```
alanshaw:
```suggestion
    return { connectionFSM, didCreate }
```

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+  /**
+   * Executes the next dial in the queue for the given peer
+   * @private
+   * @param {PeerInfo} peerInfo
+   * @returns {void}
+   */
+  _run (peerInfo) {
+    const id = peerInfo.id.toB58String()
+
+    // If we have no items in the queue, exit
```
alanshaw: also...don't continue if the queue has already been stopped?

src/connection/index.js
```diff
@@ -272,22 +272,35 @@ class ConnectionFSM extends BaseConnection {
...
+          resolve()
+        }).catch(() => { })
+      )
+    }
+
+    Promise.all(tasks)
+      .then(() => {
         this._state('done')
-        delete this.conn
       })
```
alanshaw: IMHO we should only mix promises and callbacks when we have to deal with an external library we have no control over. We already have `async` as a dependency so we'd not be saving any bundle bytes by using builtins. I think it sends the wrong message to contributors. We have a awesome endeavour where we're trying to coordinate all of this _at the same time_ and adding Promises here piecemeal is problematic for a number of reasons:

1. It's super inconsistent with the rest of the libp2p code (everything is callbacks, not even a dual promise/callback api)
1. Sets the precedent for other contributors to do the same thing, when actually we want contributions to either focus on fixing problems **or** focus on the endeavour
1. Finally, without going into specifics, mixing callbacks and Promises is fraught with double callback danger, perhaps not in this specific case, but it can be super hard to debug when it happens!

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+   * such as: /echo/1.0.0, /ipfs/kad/1.0.0, etc.
+   *
+   * @private
+   * @static
+   * @param {ConnectionFSM} connection
+   * @returns {Boolean}
+   */
+  static canShake (connection) {
+    return connection && (connection.getState() === 'MUXED' || connection.getState() === 'CONNECTED')
+  }
```
alanshaw: Nit: The existence check is redundant in the case where you're using this

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+   * @private
+   * @static
+   * @param {object} options
+   * @param {string} options.protocol
+   * @param {Connection} options.proxyConnection
+   * @param {ConnectionFSM} options.connection
+   * @param {function(Error, Connection)} options.callback
+   * @returns {void}
+   */
+  static getStreamForProtocol ({ protocol, proxyConnection, connection, callback }) {
```
alanshaw: Suggestion: rename and make private? The return value is not used and this has a side effect of updating the passed proxy.

src/dialer/queue.js
```diff
@@ -0,0 +1,253 @@
...
+   * Checks to see if the provided `connection` is capable of
+   * performing a handshake with an application level protocol,
+   * such as: /echo/1.0.0, /ipfs/kad/1.0.0, etc.
+   *
+   * @private
+   * @static
+   * @param {ConnectionFSM} connection
+   * @returns {Boolean}
+   */
+  static canShake (connection) {
```
alanshaw: Does this need to be part of the public API for this class?

It's so small and only used once, I'd be tempted to just inline it:

```js
const canShake = ['MUXED', 'CONNECTED'].includes(connection.getState())
```

src/get-peer-info.js
```diff
@@ -38,7 +38,7 @@ function getPeerInfo (peer, peerBook) {
     try {
       return peerBook.get(peerIdB58Str)
     } catch (err) {
-      throw new Error('Couldnt get PeerInfo')
+      throw new Error(`Couldnt get PeerInfo for ${peerIdB58Str}`)
```
alanshaw, line 41:
```suggestion
      throw new Error(`Failed to get PeerInfo for ${peerIdB58Str}`)
```

alanshaw (changes requested):
This looks good 👍 - most of my comments are around code organisation so pretty much all suggestions really.

src/connection/handler.js
jacobheun: The diff is just making it look like that I think, they haven't change places. `protect()` always initiates the connection flow.

alanshaw: oh, yes you're right!

src/connection/index.js
jacobheun: agreed, will move this to `async/parallel`

src/connection/index.js
jacobheun: This is now controlled by the dialer queue.

src/dialer/queue.js
jacobheun: This is due to switch being a reserved word. In other places incorrect spellings are used, which I am not a fan of as I it tends to lead to typo errors. The `_` is confusing though, as it can also indicate an unused parameter.  I think we can look at this in a subsequent PR to clean that up everywhere.

jacobheun:
Aside from the `_switch` param changes, the rest of the feedback should be implemented.

src/dialer/queue.js
alanshaw: Ah, that makes sense - thanks!

alanshaw (approved):
LGTM

PR #312, opened 2019-03-21
fix: ensure dials always use the latest PeerInfo from the PeerBook

jacobheun:
This fixes an issue where if dial is called with a new instance of PeerInfo, if it is the first dial to that peer, the queue was forever associated with that instance. This is currently the case when Circuit checks the HOP status of a potential relay. This ensures that whenever we dial, we are updating the peer book and using the latest PeerInfo in that dial request.

Fixes the issues here https://github.com/libp2p/js-libp2p/pull/336/checks?check_run_id=82068526, passing build at https://travis-ci.com/libp2p/js-libp2p/builds/105269120

### DIFF
--- a/src/dialer/queue.js
+++ b/src/dialer/queue.js
@@ -65,7 +65,6 @@ class Queue {
    * @param {Switch} _switch
    */
   constructor (peerInfo, _switch) {
-    this.peerInfo = peerInfo
     this.id = peerInfo.id.toB58String()
     this.switch = _switch
     this._queue = []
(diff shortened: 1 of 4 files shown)


src/dialer/queue.js
```diff
@@ -65,7 +65,6 @@ class Queue {
    * @param {Switch} _switch
    */
   constructor (peerInfo, _switch) {
-    this.peerInfo = peerInfo
```
alanshaw, line 68: Just pass Peer ID instead?

alanshaw (changes requested):

src/dialer/queue.js
jacobheun: fixed

alanshaw (approved):

PR #313, opened 2019-03-22
fix: identify on dial

jacobheun:
Previously identify was only run by the receiver. This change is inline with how go operates.

Required by https://github.com/libp2p/js-libp2p-daemon/pull/14
- [x] Requires https://github.com/libp2p/js-libp2p-spdy/pull/86


vasco-santos (commented):
This looks good! Needs just the new release of `libp2p-spdy` to be ready

vasco-santos (approved):

PR #314, opened 2019-03-27
feat: global dial queue

jacobheun:
This adds a general dial queue to the switch. All dial requests are added to the QueueManager. The QueueManager ensures no more than the maximum parallel dials requests are allowed (currently 50).

This also adds basic blacklisting with a ttl of 120 seconds. If a dial attempt errors, no dials to it will be allowed until the ttl has expired, or the blacklist has been manually cleared. Future iterations of this should add an exponential backoff for blacklisting to better manage repeat offending peers.

This alleviates issues where libp2p discovers a large number of peers, such as through the dht, and dial attempts are made to all of them. High numbers of concurrent dials lead to the cpu overloading.

### DIFF
--- a/src/dialer/queue.js
+++ b/src/dialer/queue.js
@@ -230,8 +230,8 @@ class Queue {
     // In the future, it may be desired to error the other queued dials,
     // depending on the error.
     connectionFSM.once('error', (err) => {
-      this.blacklist()
       queuedDial.callback(err)
+      this.blacklist()
     })
 
     connectionFSM.once('close', () => {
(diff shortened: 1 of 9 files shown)


src/dialer/queue.js
```diff
@@ -188,6 +230,7 @@ class Queue {
     // In the future, it may be desired to error the other queued dials,
     // depending on the error.
     connectionFSM.once('error', (err) => {
+      this.blacklist()
       queuedDial.callback(err)
```
alanshaw, line 233: Call the callback for the dial and then blacklist?

alanshaw (approved):
Just one suggestion :) This'll do the job for now until we can improve perf or assign some sort of priority to connections.

vasco-santos (approved):
👍

PR #315, opened 2019-03-28
fix: improve dial queue and parallel dials

jacobheun:
## Improved Dialer
This improves the dialer queue a bit by doing a few things:

**Multidimensional queues**
As dial requests are made they no longer go into a single, global queue, they are immediately added to the dial queue for that specific peer. This allows a successful queue, or one that errors, to immediately resolve the remaining dialing requests. Previously, a subsequent request would need to wait until its turn globally, which is slow, especially if it just needs to be aborted or a new stream needs to be created for a protocol.

**Immediate dials for connected peers**
If a peer is already connected, we immediately start its queue if it's not already running. The max parallel queues aren't affected by this. This ensures that we have a steady stream of communication with our connected peers, which is critical to maintaining a healthy network.

## Configurable dial timeout
The dial timeout was also not configurable. This has been corrected and added to the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,7 @@ If defined, `options` should be an object with the following keys and respective
 
 - `blacklistTTL`: - number of ms a peer should not be dialable to after it errors. Defaults to `120000`(120 seconds)
 - `maxParallelDials`: - number of concurrent dials the switch should allow. Defaults to `50`
-- `dialTimeout`: - number of ms a dial to a peer should be allowed to run. Defaultd to `30000` (30 seconds)
+- `dialTimeout`: - number of ms a dial to a peer should be allowed to run. Defaults to `30000` (30 seconds)
 - `stats`: an object with the following keys and respective values:
   - `maxOldPeersRetention`: maximum old peers retention. For when peers disconnect and keeping the stats around in case they reconnect. Defaults to `100`.
   - `computeThrottleMaxQueueSize`: maximum queue size to perform stats computation throttling. Defaults to `1000`.
--- a/README.md
+++ b/README.md
@@ -62,7 +62,8 @@ const sw = new switch(peerInfo , peerBook [, options])
 If defined, `options` should be an object with the following keys and respective values:
 
 - `blacklistTTL`: - number of ms a peer should not be dialable to after it errors. Defaults to `120000`(120 seconds)
-- `maxParallelDials` - number of concurrent dials the switch should allow. Defaults to `50`
+- `maxParallelDials`: - number of concurrent dials the switch should allow. Defaults to `50`
+- `dialTimeout`: - number of ms a dial to a peer should be allowed to run. Defaultd to `30000` (30 seconds)
 - `stats`: an object with the following keys and respective values:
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -62,7 +62,8 @@ const sw = new switch(peerInfo , peerBook [, options])
 If defined, `options` should be an object with the following keys and respective values:
 
 - `blacklistTTL`: - number of ms a peer should not be dialable to after it errors. Defaults to `120000`(120 seconds)
-- `maxParallelDials` - number of concurrent dials the switch should allow. Defaults to `50`
+- `maxParallelDials`: - number of concurrent dials the switch should allow. Defaults to `50`
+- `dialTimeout`: - number of ms a dial to a peer should be allowed to run. Defaultd to `30000` (30 seconds)
```
alanshaw:
```suggestion
- `dialTimeout`: - number of ms a dial to a peer should be allowed to run. Defaults to `30000` (30 seconds)
```

src/limit-dialer/index.js
```diff
@@ -41,20 +41,27 @@ class LimitDialer {
...
+          successfulDial = res
+        }
+
+        cb()
       })
     })
 
-    tryEach(tasks, (_, result) => {
-      if (result && result.conn) {
+    parallel(tasks, (_) => {
```
alanshaw: Use `async/each` instead so you can share the task function?

```js
each(addrs, (m, cb) => { this.dialSingle(/*...*/) })
```

alanshaw: Actually, do you want to use `async/race` for this instead so you don't have to wait for all the dials to finish before calling back?

src/limit-dialer/queue.js
```diff
@@ -54,9 +54,9 @@ class DialQueue {
         pull(empty(), conn)
         // If we can close the connection, do it
         if (typeof conn.close === 'function') {
-          return conn.close((_) => callback(null, { cancel: true }))
+          return conn.close((_) => callback(null))
         }
-        return callback(null, { cancel: true })
+        return callback(null)
```
alanshaw, line 59: What does removing these objects do?

alanshaw (changes requested):

src/limit-dialer/queue.js
jacobheun: I believe these were left over from some old code, only valid connection results are being used now, so they were deleted. The intent is that it creates a cancel token that causes other successfully dialed connections to be destroyed. Now the token is created when we receive the first valid connection.

src/limit-dialer/index.js
jacobheun: I will use each. `race` will finish as soon as any of the dial callbacks finish, even if nothing is passed. So if we dial 5 addresses and the first errors, even if we just store an array of the errors as soon as we callback it ends, so we end up having a "failed" dial after the first one errors. Even if the other 4 were actually good.

jacobheun: Okay, I did go with race for the aforementioned benefits. I just only call the callback if we have a value, or if everything errored.

alanshaw (approved):

PR #316, opened 2019-04-01
feat: limit the number of cold calls we can do

jacobheun:
This creates a separate queue for "cold calls" (dials without a protocol). Applications typically do this on peer discovery to connect to new peers. This is to prevent flooding the normal queue, which needs to be prioritized for normal operations, with dials to new peers that may not respond.

This is a workaround, until we can improve the peer discovery system and connection/peer tagging, to allow for proper priority. Right now all dials are treated the same, which includes just creating a new stream on an existing connection. Existing connections and new dials should not be throttled the same way. 

This also adds a very basic backoff to blacklisted peers. On its 5th blacklisting the peer will be added to the blacklist permanently. *The backoff time may need to be increased before merging this*

### DIFF
--- a/src/dialer/queue.js
+++ b/src/dialer/queue.js
@@ -152,12 +152,19 @@ class Queue {
   blacklist () {
     this.blackListCount++
 
-    if (this.blackListCount >= 5) {
+    if (this.blackListCount >= this.switch.dialer.BLACK_LIST_ATTEMPTS) {
       this.blackListed = Infinity
       return
     }
 
-    this.blackListed = Date.now() + (this.switch.dialer.BLACK_LIST_TTL * Math.pow(this.blackListCount, 3))
+    let ttl = this.switch.dialer.BLACK_LIST_TTL * Math.pow(this.blackListCount, 3)
+    const minTTL = ttl * 0.9
+    const maxTTL = ttl * 1.1
+
+    // Add a random jitter of 20% to the ttl
+    ttl = Math.floor(Math.random() * (maxTTL - minTTL) + minTTL)
+
+    this.blackListed = Date.now() + ttl
     this.abort()
   }
 
@@ -253,6 +260,7 @@ class Queue {
 
     connectionFSM.once('unmuxed', () => {
       this.blackListCount = 0
+      this.switch.connection.add(connectionFSM)
       queuedDial.connection = connectionFSM
       createConnectionWithProtocol(queuedDial)
       next()
--- a/src/dialer/queue.js
+++ b/src/dialer/queue.js
@@ -157,7 +157,7 @@ class Queue {
       return
     }
 
-    this.blackListed = Date.now() + (this.switch.dialer.BLACK_LIST_TTL * Math.pow(this.blackListCount, 2))
+    this.blackListed = Date.now() + (this.switch.dialer.BLACK_LIST_TTL * Math.pow(this.blackListCount, 3))
     this.abort()
   }
 
--- a/src/dialer/queue.js
+++ b/src/dialer/queue.js
@@ -70,6 +70,7 @@ class Queue {
     this.switch = _switch
     this._queue = []
     this.blackListed = null
+    this.blackListCount = 0
     this.isRunning = false
     this.onStopped = onStopped
   }
@@ -97,7 +98,7 @@ class Queue {
   isDialAllowed () {
     if (this.blackListed) {
       // If the blacklist ttl has passed, reset it
-      if (Date.now() - this.blackListed > this.switch.dialer.BLACK_LIST_TTL) {
+      if (Date.now() > this.blackListed) {
         this.blackListed = null
         return true
       }
@@ -146,10 +147,17 @@ class Queue {
 
   /**
    * Marks the queue as blacklisted. The queue will be immediately aborted.
+   * @returns {void}
    */
   blacklist () {
-    log('blacklisting queue for %s', this.id)
-    this.blackListed = Date.now()
+    this.blackListCount++
+
+    if (this.blackListCount >= 5) {
+      this.blackListed = Infinity
+      return
+    }
+
+    this.blackListed = Date.now() + (this.switch.dialer.BLACK_LIST_TTL * Math.pow(this.blackListCount, 2))
     this.abort()
   }
 
--- a/src/dialer/queue.js
+++ b/src/dialer/queue.js
@@ -244,6 +244,7 @@ class Queue {
 
     // If we're not muxed yet, add listeners
     connectionFSM.once('muxed', () => {
+      this.blackListCount = 0 // reset blacklisting on good connections
       this.switch.connection.add(connectionFSM)
       queuedDial.connection = connectionFSM
       createConnectionWithProtocol(queuedDial)
@@ -251,6 +252,7 @@ class Queue {
     })
 
     connectionFSM.once('unmuxed', () => {
+      this.blackListCount = 0
       queuedDial.connection = connectionFSM
       createConnectionWithProtocol(queuedDial)
       next()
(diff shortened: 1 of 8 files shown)


src/dialer/queue.js
```diff
@@ -146,10 +147,17 @@ class Queue {
...
   /**
    * Marks the queue as blacklisted. The queue will be immediately aborted.
+   * @returns {void}
    */
   blacklist () {
-    log('blacklisting queue for %s', this.id)
-    this.blackListed = Date.now()
+    this.blackListCount++
+
+    if (this.blackListCount >= 5) {
```
dirkmc: Should this number be configurable?

src/dialer/queue.js
```diff
@@ -146,10 +147,17 @@ class Queue {
...
+      this.blackListed = Infinity
```
dirkmc, line 156: Infinity seems like a long time :)

I'm thinking of the case where a peer goes offline for a couple of hours and then comes back online.

src/dialer/queue.js
```diff
@@ -236,13 +244,15 @@ class Queue {
 
     // If we're not muxed yet, add listeners
     connectionFSM.once('muxed', () => {
+      this.blackListCount = 0 // reset blacklisting on good connections
```
dirkmc, line 254: Should `this.blackListed` also be reset?

src/dialer/queueManager.js
```diff
@@ -54,10 +62,18 @@ class DialQueueManager {
...
-    // and if the queue is allowed to dial
-    if (!targetQueue.isRunning && targetQueue.isDialAllowed()) {
-      this._queue.add(targetQueue.id)
+    // If dialing is not allowed, abort
+    if (!targetQueue.isDialAllowed()) {
+      return
+    }
+
+    // Add the id to its respective queue set if the queue isn't running
+    if (!targetQueue.isRunning) {
```
dirkmc, line 121: Should it be possible for a peer to be added to both queues? If not we may need to check here that it's not already in the other queue.

src/dialer/queue.js
jacobheun: yes! it now is and has a default of 5

src/dialer/queue.js
jacobheun: it sure is! Really what I'm trying to do is get a reasonable, increasing backoff. Right now, not including the jitter i just added, it caps at ~5.3 hours (~8.3 hours if you take into account the successive failures it would take to get there). Perhaps this scale isn't sufficient. (5min, 40min, 135min, 320min).

I do think we could scale it back from Infinity though, perhaps a max of 1day would be large enough. This would basically give us the processing of: if I can't dial you over the course of 8 hours, I'll just wait until tomorrow. Each failure after that would also postpone a day. 

Eventually a "dead" node should stop getting emitted as a peer.

src/dialer/queue.js
jacobheun: If we're able to dial the peer it should already be null, since `this.blackListed ` is checked before any dial attempt is made. Setting this to 0 ensures that if we weren't able to previously dial the peer that we clear the blacklist so we don't end up waiting a day to dial if it has some intermittent availability.

src/dialer/queue.js
dirkmc: Having a back-off makes sense 👍 

Maybe we could have another process that removes peers from the blacklist after a certain amount of time so as to limit memory usage. It could be an issue for a node that sees a ton of ephemeral peers (eg a gateway node).

~~Is there a mechanism by which peers are removed from the blacklist on discovery?~~ nvm I see they get removed on muxed

src/dialer/queueManager.js
jacobheun: We should avoid them being in both. I'll update the logic to check that.

jacobheun: Added some checks and tests to ensure it's only in 1 queue

src/dialer/queue.js
jacobheun: What if we had the queue manager go through its queues on a daily/semi-daily(?) interval and had it simply delete any queues with an `Infinity` blacklist? If the peers are discovered again, we could attempt to dial them, and if not we free up the memory of that queue. That might also be a good trigger to remove it from `PeerBook`.

dirkmc: That sounds like a reasonable approach. I would favour frequent garbage collection, as it's not a very expensive operation and freeing up memory could be critical for limited memory environments. We might also want to be a config option for a maximum size for the queue.

src/dialer/queueManager.js
```diff
@@ -34,7 +37,7 @@ class DialQueueManager {
   }
 
   /**
-   * Adds the `dialRequest` to the queue and ensures the queue is running
+   * Adds the `dialRequest` to the qMAX_COLD_CALLSe queue is running
```
dirkmc: wut? :)

jacobheun: bah, woops

jacobheun:
Found an unrelated connection management issue, https://github.com/libp2p/js-libp2p-switch/issues/317,  as I was running this against my local js-ipfs daemon logging out some stats.

src/dialer/queue.js
jacobheun: I added an hourly timer to abort and remove queues that have hit a blacklist of `Infinity`. 

> We might also want to be a config option for a maximum size for the queue.

Which queue are you referring to? `_queue`? If so, Ideally we'd avoid setting a max size on this. If we can't get through `_queue` quickly that means our normal node operations are going to suffer. So far my testing locally, has seen `_queue` relatively low and it clears fairly quickly. The cold queue is usually close to max, except right at startup.

dirkmc: Sorry I meant the list of dial queues in QueueManager. It just keeps growing in size, right?

jacobheun: Yes, there's currently a DialQueue for every peer we've attempted to connect with, unless it's cleaned via the hourly cleanup.

I guess we do need to do some cleanup there. Queues that aren't active and aren't tracking blacklisting shouldn't need to stay around. Ideally I think blacklisting info would be stored as metadata in PeerBook, this would prevent us holding onto queues unnecessarily.

jacobheun: I reduced the cleanup time down to 15mins and added some more logic there. If it's not connected, not blacklisted and not queued to be run, we get rid of it. If it matches those other criteria we keep it to avoid frequently recreating the queues.

vasco-santos: I liked the way you got this!

src/dialer/queueManager.js
```diff
@@ -11,9 +15,53 @@ class DialQueueManager {
...
+   * is considered no longer valid.
+   * @private
+   */
+  _clean () {
+    const queues = Object.values(this._queues)
+    queues.forEach(dialQueue => {
+      // Clear if the queue has reached max blacklist
+      if (dialQueue.blackListed === Infinity) {
+        dialQueue.abort()
+        delete this._queues[dialQueue.id]
```
vasco-santos, line 37: What do you think about doing `this._queues[dialQueue.id] = undefined`?

This is more efficient than deleting, but than we will need to verify that keys have a value when iterating the array, so I do not have a strong opinion here.

vasco-santos (approved):
Looks great @jacobheun ! Nice work

Approved, but left a minor suggestion to consider

src/dialer/queueManager.js
jacobheun: Since this is only running on a 15min interval I think the performance hit should be fairly insignificant and would avoid us needing to check values.

dirkmc:
LGTM 👍

PR #318, opened 2019-04-02
fix: improve connection tracking

jacobheun:
Previously, `peer-mux-established` and `peer-muxed-closed` events were being fired anytime a connection was created or closed. The problem with this is that if you happened to have 2 connections to a peer and only 1 of them closed, the closed event would fire, causing the connection manager to believe that peer no longer had any connections. This would result in the connection manager believing there were way fewer connections than actually existed. 

This PR moves the emit events and `PeerInfo.disconnect` calls into the `switch.connection` class. When connections are added and removed it will check to see if there are any additional connections to that peer. This way, we only fire events when changes actually occur.

### DIFF
--- a/src/connection/incoming.js
+++ b/src/connection/incoming.js
@@ -68,9 +68,6 @@ class IncomingConnectionFSM extends BaseConnection {
       this.emit('muxed', this.conn)
     })
     this._state.on('DISCONNECTING', () => {
-      if (this.theirPeerInfo) {
-        this.theirPeerInfo.disconnect()
-      }
       this._state('done')
     })
   }
--- a/src/connection/index.js
+++ b/src/connection/index.js
@@ -268,11 +268,6 @@ class ConnectionFSM extends BaseConnection {
   _onDisconnecting () {
     this.log('disconnecting from %s', this.theirB58Id, Boolean(this.muxer))
 
-    // Issue disconnects on both Peers
-    if (this.theirPeerInfo) {
-      this.theirPeerInfo.disconnect()
-    }
-
     this.switch.connection.remove(this)
 
     delete this.switch.conns[this.theirB58Id]
@@ -284,7 +279,6 @@ class ConnectionFSM extends BaseConnection {
       tasks.push((cb) => {
         this.muxer.end(() => {
           delete this.muxer
-          this.switch.emit('peer-mux-closed', this.theirPeerInfo)
           cb()
         })
       })
@@ -325,13 +319,13 @@ class ConnectionFSM extends BaseConnection {
           return this.close(maybeUnexpectedEnd(err))
         }
 
-        const conn = observeConnection(null, this.switch.crypto.tag, _conn, this.switch.observer)
-
-        this.conn = this.switch.crypto.encrypt(this.ourPeerInfo.id, conn, this.theirPeerInfo.id, (err) => {
+        const observedConn = observeConnection(null, this.switch.crypto.tag, _conn, this.switch.observer)
+        const encryptedConn = this.switch.crypto.encrypt(this.ourPeerInfo.id, observedConn, this.theirPeerInfo.id, (err) => {
           if (err) {
             return this.close(err)
           }
 
+          this.conn = encryptedConn
           this.conn.setPeerInfo(this.theirPeerInfo)
           this._state('done')
         })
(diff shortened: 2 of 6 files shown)


src/connection/incoming.js
```diff
@@ -68,9 +68,6 @@ class IncomingConnectionFSM extends BaseConnection {
       this.emit('muxed', this.conn)
     })
     this._state.on('DISCONNECTING', () => {
-      if (this.theirPeerInfo) {
-        this.theirPeerInfo.disconnect()
-      }
```
jacobheun, line 73: This was unnecessary. Incoming connections don't cause the connect to happen until after the are muxed, which will now be handled by the `switch.connection.add()` logic.

src/connection/index.js
```diff
@@ -325,13 +319,13 @@ class ConnectionFSM extends BaseConnection {
           return this.close(maybeUnexpectedEnd(err))
         }
 
-        const conn = observeConnection(null, this.switch.crypto.tag, _conn, this.switch.observer)
-
-        this.conn = this.switch.crypto.encrypt(this.ourPeerInfo.id, conn, this.theirPeerInfo.id, (err) => {
```
jacobheun, line 330: I found an edge case here, where if the connection was terminated while encryption was being negotiated, `this.conn` would get resolved and would later throw an error when secio went to set the inner connection. This avoids that.

src/dialer/queue.js
```diff
@@ -45,10 +44,8 @@ function createConnectionWithProtocol ({ protocol, connection, callback }) {
       return callback(err)
     }
 
-    const proxyConnection = new Connection()
-    proxyConnection.setPeerInfo(connection.theirPeerInfo)
-    proxyConnection.setInnerConn(conn)
-    callback(null, proxyConnection)
+    conn.setPeerInfo(connection.theirPeerInfo)
+    callback(null, conn)
```
jacobheun, line 48: The `proxyConnection` was superfluous here, so I got rid of it.

src/connection/index.js
dirkmc: Nice catch!

src/connection/manager.js
```diff
@@ -78,14 +79,26 @@ class ConnectionManager {
...
   remove (connection) {
-    if (!this.connections[connection.theirB58Id]) return
+    // No record of the peer, disconnect it
+    if (!this.connections[connection.theirB58Id]) {
+      connection.theirPeerInfo.disconnect()
+      this.switch.emit('peer-mux-closed', connection.theirPeerInfo)
+      return
+    }
 
     for (let i = 0; i < this.connections[connection.theirB58Id].length; i++) {
```
dirkmc, line 89: I feel like the code in this class could be simpler with `Set` and `Map` - suggestion for a future PR

src/connection/manager.js
```diff
@@ -185,24 +199,24 @@ class ConnectionManager {
...
-              // of the available multiaddrs from the other peer as the one
-              // I'm connected to as we really can't be sure at the moment
-              // TODO add this consideration to the connection abstraction!
-              peerInfo.connect(peerInfo.multiaddrs.toArray()[0])
-            } else {
-              // for the case of websockets in the browser, where peers have
-              // no addr, use just their IPFS id
-              peerInfo.connect(`/ipfs/${b58Str}`)
+            // Only update if it's not already connected
+            if (!peerInfo.isConnected()) {
```
dirkmc, line 203: How reliable is peerInfo.isConnected()? I ask because it seems like things are put in the peer book from a lot of different places in the code

dirkmc:
I left a couple of questions in the code but in general it seems like a much more reliable way of doing things to me 👍

src/connection/manager.js
jacobheun: `PeerBook.put()` is definitely overused. I think clearly defining the peer discovery pipeline and responsibility around who actually should be putting would be helpful.  `PeerInfo.isConnected` has been showing questionable reliability in my local checks, but I haven't determine the root cause of that yet. I did a search of modules in js-ipfs for `.connect` calls (although I need to check `.disconnect`) to see if any modules are calling those, which they shouldn't be.

If `PeerInfo`returns a connected address we should be confident in its validity though. The main goal here is to avoid stomping over a more useful dialed address with an incoming address, as we know for a fact the former was used in a successful dial.

src/connection/manager.js
jacobheun: I agree, a `Map` would allow for handling the two dimensional array pretty cleanly.

dirkmc:
LGTM 👍

vasco-santos (approved):
LGTM!

Good to know that the inconsistency between the number of peers in the connection manager and in reality should be fixed.

PR #320, opened 2019-04-03
fix: revert to try each

alanshaw:
Race was opening multiple connections and not closing/canceling the unused ones.

### DIFF
--- a/src/limit-dialer/index.js
+++ b/src/limit-dialer/index.js
@@ -38,17 +38,25 @@ class LimitDialer {
     // we use a token to track if we want to cancel following dials
     const token = { cancel: false }
 
+    let errors = []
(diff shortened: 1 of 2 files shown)


jacobheun:
This should revert back to https://github.com/libp2p/js-libp2p-switch/blob/ddf622b784de1d3d376f9c429a49a84fbd660ca9/src/limit-dialer/index.js.  Currently only the last error will be returned from the limit dialer, whereas https://github.com/libp2p/js-libp2p-switch/blob/ddf622b784de1d3d376f9c429a49a84fbd660ca9/src/limit-dialer/index.js will callback with all errors. This is helpful for diagnosing failed dials because you can see what went wrong with each dial, instead of just the last.

PR #322, opened 2019-04-03
fix: running queues after stop

### LINKED ISSUE #321: Error on stop
```console
^CReceived interrupt signal, shutting down...
/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/queueManager.js:167
      this._dialingQueues.add(targetQueue.id)
                                          ^

TypeError: Cannot read property 'id' of undefined
    at DialQueueManager.run (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/queueManager.js:167:43)
    at DialQueueManager._onQueueStopped (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/queueManager.js:190:10)
    at Queue.stop (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/queue.js:130:12)
    at Queue.abort (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/queue.js:142:10)
    at queues.forEach.dialQueue (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/queueManager.js:94:17)
    at Array.forEach (<anonymous>)
    at DialQueueManager.abort (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/queueManager.js:93:12)
    at EventEmitter.abort (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/libp2p-switch/src/dialer/index.js:44:22)
    at EventEmitter.emit (events.js:193:13)
    at enter (/Users/alan/Code/protocol-labs/js-ipfs/node_modules/fsm-event/index.js:57:20)
```

alanshaw:
resolves #321

### DIFF
--- a/src/dialer/index.js
+++ b/src/dialer/index.js
@@ -12,7 +12,8 @@ const {
 module.exports = function (_switch) {
   const dialQueueManager = new DialQueueManager(_switch)
 
-  _switch.state.on('STOPPING:enter', abort)
+  _switch.state.on('STARTING:enter', start)
+  _switch.state.on('STOPPING:enter', stop)
 
   /**
    * @param {DialRequest} dialRequest
@@ -35,13 +36,23 @@ module.exports = function (_switch) {
   }
 
   /**
+   * Signals to the dialer that it should start processing dial queues
+   *
+   * @param {function} callback
+   */
+  function start (callback) {
+    dialQueueManager.start()
+    callback()
+  }
+
+  /**
    * Aborts all dials that are queued. This should
    * only be used when the Switch is being stopped
    *
    * @param {function} callback
    */
-  function abort (callback) {
-    dialQueueManager.abort()
+  function stop (callback) {
+    dialQueueManager.stop()
     callback()
   }
 
@@ -77,7 +88,8 @@ module.exports = function (_switch) {
   return {
     dial,
     dialFSM,
-    abort,
+    start,
+    stop,
     clearBlacklist,
     BLACK_LIST_ATTEMPTS: isNaN(_switch._options.blackListAttempts) ? BLACK_LIST_ATTEMPTS : _switch._options.blackListAttempts,
     BLACK_LIST_TTL: isNaN(_switch._options.blacklistTTL) ? BLACK_LIST_TTL : _switch._options.blacklistTTL,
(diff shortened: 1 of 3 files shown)


jacobheun:
The issue with this current approach is that the switch doesn't currently prohibit dials when it is stopped (which is why some tests are failing, they never start the switch). While libp2p currently prohibits dials when it's offline, it may be something that warrants changing in the future, https://github.com/libp2p/js-libp2p/issues/341.

It might be simpler to just ensure the target queue exists before adding it and starting it, to avoid the change in switch behavior. This is likely an edge case with running queues shutting down while the queues are draining.

alanshaw:
...but if the switch is stopped isn't it reasonable for it to disallow dials? I take the point on https://github.com/libp2p/js-libp2p/issues/341 but should there instead be different modes of operation? e.g. "start in dial only mode"

My issue is that https://github.com/libp2p/js-libp2p-switch/pull/323 will guard against the edge case but it doesn't actually solve the issue of the node continuing to run after it has been "stopped".

jacobheun:
> My issue is that #323 will guard against the edge case but it doesn't actually solve the issue of the node continuing to run after it has been "stopped".

Yeah sorry, I confused the goal being to just fix #323. I am looking into an issue where there are connections staying open after close, even when the queues are stopped which is likely contributing to this problem.

I do think a configuration of dialOnly or offline is the best route to go for supporting libp2p/js-libp2p#341 properly.

alanshaw:
> Yeah sorry, I confused the goal being to just fix #323.

I should have made it clearer. Do you think what's here is valuable? If so, I'm happy to do the work the fix up the tests.

jacobheun:
> Do you think what's here is valuable?

Yes, what if we had the dial manager initialize with `this._isRunning` set to true instead of false? This would keep the current functionality of allowing offline dials, but once start/stop actions are taken it's no longer considered offline and the dial manager will behave as such.

alanshaw:
Superceded by #324

PR #325, opened 2019-04-09
feat: support a priority queue for dials

jacobheun:
This adds basic support for priorities. Currently, all priority 0 (highest) items will be run through the main dial queue and all priorities >= 1 will go through the cold call queue. This enables the proposed auto dial feature in libp2p, https://github.com/libp2p/js-libp2p/pull/349, to create low priority dials on peer discovery.

Since some applications may dial to peers without specifying a protocol, such as dialing to a content provider discovered through the DHT, only detecting the presence of a protocol is not sufficient to determine priority. `switch.dialer.connect` creates a direct path for applications to use the cold call queue. 

required by https://github.com/libp2p/js-libp2p/pull/349

### DIFF
--- a/README.md
+++ b/README.md
@@ -154,13 +154,11 @@ and to received the underlying connection that can be used to transfer data.
 
 ### `switch.dialer.connect(peer, options, callback)`
 
-a low priority dial to the provided peer. Calls to `dial` and `dialFSM` will take priority. This should be used when
-an application only wishes to establish connections to new peers, such as during peer discovery when there is a low
-peer count.
+a low priority dial to the provided peer. Calls to `dial` and `dialFSM` will take priority. This should be used when an application only wishes to establish connections to new peers, such as during peer discovery when there is a low peer count. Currently, anything greater than the HIGH_PRIORITY (10) will be placed into the cold call queue, and anything less than or equal to the HIGH_PRIORITY will be added to the normal queue.
 
 - `peer`: can be an instance of [PeerInfo][], [PeerId][] or [multiaddr][]
 - `options`: Optional
-- `options.priority`: Number of the priority of the dial, defaults to 1.
+- `options.priority`: Number of the priority of the dial, defaults to 20.
 - `options.useFSM`: Boolean of whether or not to callback with a [Connection State Machine](#connection-state-machine)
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -152,6 +152,18 @@ works like dial, but calls back with a [Connection State Machine](#connection-st
...
 
+### `switch.dialer.connect(peer, options, callback)`
+
+a low priority dial to the provided peer. Calls to `dial` and `dialFSM` will take priority. This should be used when
+an application only wishes to establish connections to new peers, such as during peer discovery when there is a low
+peer count.
+
+- `peer`: can be an instance of [PeerInfo][], [PeerId][] or [multiaddr][]
+- `options`: Optional
+- `options.priority`: Number of the priority of the dial, defaults to 1.
```
vasco-santos: What is the acceptable range?

src/dialer/index.js
```diff
@@ -64,14 +64,32 @@ module.exports = function (_switch) {
...
+    if (typeof options === 'function') {
+      callback = options
+      options = null
+    }
+    options = { useFSM: false, priority: 1, ...options }
+    _dial({ peerInfo, protocol: null, options, callback })
+  }
+
   /**
    * Adds the dial request to the queue for the given `peerInfo`
```
vasco-santos, line 88: Add docs for using higher priority here

README.md
dirkmc: Should we start with higher numbered priorities to leave some wiggle room in case there are other use cases in future? eg HIGH=10, LOW=20, and then in future maybe there's a MEDIUM=15

dirkmc (approved):
One small suggestion, otherwise LGTM 👍

README.md
jacobheun: I can't think of anything now, but that doesn't mean there wont be :). I've added those as constants to the constants file, and am now referencing them for the options.

vasco-santos (approved):
Looks good to me! Only left 2 minor comments

README.md
vasco-santos: oh ups, I did not see this review discussion and left a comment in this regard too

jacobheun: Since items currently go into 1 of 2 queues that execute by order of adds, I updated the docs to try to be more clear about what's going on. I think if we expand on that in the future or actually add priority to each of the queues we can lock down what's acceptable.

PR #328, opened 2019-04-10
fix: create separate events for connection tracking

jacobheun:
The `peer-mux-*` events are currently used for both connection tracking and muxed connection tracking. This creates a problem for the Connection Manager because it won't be able to track unmuxed connections. 

The Connection Manager should be aware of **all** connections that are created and closed. This adds events for libp2p and subsequently the Connection Manager to hook into.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,17 @@ Emitted when the switch encounters an error.
 
 ### `switch.on('peer-mux-closed', (peer) => {})`
 
-- `peer`: is instance of [PeerInfo][] that has info of the peer we have just closed a muxed connection.
+- `peer`: is instance of [PeerInfo][] that has info of the peer we have just closed a muxed connection with.
+
+### `switch.on('connection:start', (peer) => {})`
+This will be triggered anytime a new connection, regardless of state, is added to the switches internal connection tracking.
+
+- `peer`: is instance of [PeerInfo][] that has info of the peer we have just started a connection with.
+
+### `switch.on('connection:end', (peer) => {})`
+This will be triggered anytime an existing connection, regardless of state, is removed from the switches internal connection tracking.
+
+- `peer`: is instance of [PeerInfo][] that has info of the peer we have just closed a connection with.
 
 ### `switch.on('start', () => {})`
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -198,7 +198,17 @@ Emitted when the switch encounters an error.
 
 ### `switch.on('peer-mux-closed', (peer) => {})`
 
-- `peer`: is instance of [PeerInfo][] that has info of the peer we have just closed a muxed connection.
+- `peer`: is instance of [PeerInfo][] that has info of the peer we have just closed a muxed connection with.
+
+### `switch.on('connection:start', (peer) => {})`
+This will be triggered anytime a new connection, regardless of state, is added to the switches internal connection tracking.
```
dirkmc:
```suggestion
This will be triggered anytime a new connection, regardless of state, is added to the switch's internal connection tracking.
```

README.md
```diff
@@ -198,7 +198,17 @@ Emitted when the switch encounters an error.
...
+
+- `peer`: is instance of [PeerInfo][] that has info of the peer we have just started a connection with.
+
+### `switch.on('connection:end', (peer) => {})`
+This will be triggered anytime an existing connection, regardless of state, is removed from the switches internal connection tracking.
```
dirkmc:
```suggestion
This will be triggered anytime an existing connection, regardless of state, is removed from the switch's internal connection tracking.
```

README.md
dirkmc: I'm not sure it's so clear to a user of the library what adding / removing a connection to the switch's internal connection tracking means

src/connection/manager.js
```diff
@@ -99,6 +100,9 @@ class ConnectionManager {
       connection.theirPeerInfo.disconnect()
       this.switch.emit('peer-mux-closed', connection.theirPeerInfo)
```
dirkmc, line 107: Should we remove the listener added to the connection above with `connection.once('muxed', ...)`?

dirkmc (changes requested):
Seems like the right thing to do 👍 

I left a couple of comments in the code

src/connection/manager.js
jacobheun: The connections will remove all listeners attached to them once they hit the 'disconnect' state. This is to try to and keep listener management as simple as possible.

README.md
jacobheun: This could probably just read: `This will be triggered anytime a new connection is created`

src/connection/manager.js
dirkmc: Cool that makes sense

dirkmc (approved):

PR #329, opened 2019-04-10
fix: dial self

### LINKED ISSUE #326: Switch allows dial itself
Discovery modules can emit `PeerInfo` objects with multiaddrs that **do not** have the peer ID encapsulated.

A libp2p node's own peer info seems to have multiaddrs that **do** have the peer ID encapsulated.

It means that a libp2p node will attempt to dial itself because the following test will fail, when attempting to `decapsulate` `/ip4/127.0.0.1/tcp/4002` from the peer's own address `/ip4/127.0.0.1/tcp/4002/ipfs/QmebzNV1kSzLfaYpSZdShuiABNUxoKT1vJmCdxM2iWsM2j`:

https://github.com/libp2p/js-libp2p-switch/blob/828e685e79a918831a6dac0a868d0b209dd5874c/src/transport.js#L212-L220

refs https://github.com/libp2p/js-libp2p/issues/323

alanshaw:
This PR fixes multiple cases where a node would dial itself.

fixes #326

### DIFF
--- a/src/transport.js
+++ b/src/transport.js
@@ -203,23 +203,61 @@ class TransportManager {
     if (!transport || !transport.filter) return []
 
     const transportAddrs = transport.filter(multiaddrs)
-    if (!peerInfo) {
+    if (!peerInfo || !transportAddrs.length) {
       return transportAddrs
     }
 
-    const ourAddrs = peerInfo.multiaddrs.toArray()
-    return transportAddrs.filter((addr) => {
+    const ourAddrs = ourAddresses(peerInfo)
+
+    const result = transportAddrs.filter(transportAddr => {
       // If our address is in the destination address, filter it out
-      return !ourAddrs.find((pAddr) => {
-        try {
-          addr.decapsulate(pAddr)
-        } catch (err) {
-          return false
-        }
-        return true
-      })
+      return !ourAddrs.some(a => getDestination(transportAddr).startsWith(a))
     })
+
+    return result
   }
+}
+
+/**
+ * Expand addresses in peer info into array of addresses with and without peer
+ * ID suffix.
+ *
+ * @param {PeerInfo} peerInfo Our peer info object
+ * @returns {String[]}
+ */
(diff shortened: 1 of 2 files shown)


src/transport.js
```diff
@@ -203,23 +203,61 @@ class TransportManager {
...
 
     const transportAddrs = transport.filter(multiaddrs)
-    if (!peerInfo) {
+    if (!peerInfo || !transportAddrs.length) {
       return transportAddrs
     }
 
-    const ourAddrs = peerInfo.multiaddrs.toArray()
-    return transportAddrs.filter((addr) => {
+    const ourAddrs = ourAddresses(peerInfo)
```
jacobheun, line 210: I'm wondering if we should create and store this after listening is done, instead of running it before every dial. Our addresses shouldn't change, at the moment, after listening is done. It might save us quite a bit of processing when there are a lot of dials being done.

alanshaw: I guess we could. It sounds like it would add a bunch of complexity that might easily cause confusion/bugs in the in the future if/when our addresses can change. Can we do it in a separate PR if it's a bottleneck?

Really, a lot of this should go away when addrs don't automatically carry a peer ID.

vasco-santos: I am ok with addressing this in a new PR

vasco-santos (approved):
LGTM!

src/transport.js
jacobheun: > Really, a lot of this should go away when addrs don't automatically carry a peer ID.

That's true, and since it's on my list of things to get to this quarter I think we can move forward with this. The queuing should mitigate the performance impact anyway.

jacobheun (approved):

achingbrain:
This sounds a bit like the issue I described in libp2p/js-libp2p-floodsub#58 - maybe this has fixed that too

vasco-santos:
Can you check if that got fixed @achingbrain ?

PR #330, opened 2019-04-11
fix: dont compare empty strings

jacobheun:
Currently there is something in the libp2p system that ends up adding `/ipfs/{peerId}` to our PeerInfo address list. This shouldn't be happening. While this doesn't solve that issue (I'm still determining where that's happening), it fixes an issue with transport address filtering. If any item in the list of `ourAddrs` happens to be an empty string, that will cause **all** the target peers addresses to get filtered out. This avoids that by filtering out empty strings from `ourAddrs`

### DIFF
--- a/src/transport.js
+++ b/src/transport.js
@@ -236,6 +236,7 @@ function ourAddresses (peerInfo) {
         : `${addr}/ipfs/${ourPeerId}`
       return ourAddrs.concat([addr, otherAddr])
     }, [])
+    .filter(a => Boolean(a))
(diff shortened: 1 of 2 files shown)


vasco-santos (approved):

jacobheun:
So the address is actually coming from the observed addresses from identify. We're currently just blindly adding any of those to our address list. We should do some level of verification for this. I will add a note to look into this more as a part of https://github.com/libp2p/js-libp2p/issues/202. I think it's a good place to validate we're doing the right thing with addresses.

PR #333, opened 2019-04-15
fix: resolve transport sort order in browsers

jacobheun:
This fixes an issue with how browsers handle sorting. In Chrome, the existing code would always sort Circuit last. In Firefox, it does not. The existing logic doesn't take any consideration into the `b` argument of sort. The problem with this is you will get unreliable behavior because the `a` and `b` values are sorted as "equal" if `a` is not `"Circuit"`.

This PR changes the logic to always prioritize `a` if it is not `"Circuit"`, and if it is, de-prioritize it. 

The caveat of this brief sort logic is that the other transports will have different orders in different browsers. This doesn't matter to Switch, it only cares that Circuit is last. This is why the test is set to only validate that part of the order.

required by https://github.com/ipfs/interop/issues/63

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -127,9 +127,11 @@ class Switch extends EventEmitter {
 
     // Only listen on transports we actually have addresses for
     return myTransports.filter((ts) => this.transports[ts].filter(myAddrs).length > 0)
-      // push Circuit to be the last proto to be dialed
-      .sort((a) => {
-        return a === Circuit.tag ? 1 : -1
+      // push Circuit to be the last proto to be dialed, and alphabetize the others
+      .sort((a, b) => {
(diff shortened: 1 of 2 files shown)


dirkmc:
It may make testing more deterministic if the behaviour is the same across browsers (eg sort such that Circuit is last, and then by alphabetical order of transport.tag)

vasco-santos (approved):

jacobheun:
> It may make testing more deterministic if the behaviour is the same across browsers (eg sort such that Circuit is last, and then by alphabetical order of transport.tag)

That's a good point, although it **shouldn't** matter for non circuit transports, I went ahead and updated it to sort consistently just to be sure.

PR #334, opened 2019-04-16
fix: dont use peerinfo distinct

jacobheun:
replaces https://github.com/libp2p/js-peer-info/pull/78

This is a workaround, as noted in the code comments, until https://github.com/libp2p/interface-transport/issues/41 is completed and we can allow transports to do their own unique filtering.

This removes our dependency on the `PeerInfo.multiaddrs.distinct()` method, which is a patch for this same issue. This attempts to improve on the logic there by being more specific about port uniqueness, which should also allow us to implement multiple port 0 listens, https://github.com/libp2p/js-libp2p-switch/pull/227.

### DIFF
--- a/src/utils.js
+++ b/src/utils.js
@@ -49,17 +49,17 @@ module.exports.identifyDialer = (connection, cryptoPeerInfo) => {
 }
 
 /**
- * Copied from `unique-by`, https://github.com/mlmorg/unique-by.
+ * Copied (and modified) from `unique-by`, https://github.com/mlmorg/unique-by.
  * @param {Array} arr The array to get unique values for
  * @param {function(value)} getValue The function to determine what is compared
  * @returns {Array}
  */
-module.exports.uniqueBy = function (arr, getValue) {
-  var unique = []
-  var found = {}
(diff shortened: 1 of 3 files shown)


src/utils.js
```diff
@@ -47,3 +47,24 @@ module.exports.identifyDialer = (connection, cryptoPeerInfo) => {
...
 }
+
+/**
+ * Copied (and modified) from `unique-by`, https://github.com/mlmorg/unique-by.
+ * @param {Array} arr The array to get unique values for
+ * @param {function(value)} getValue The function to determine what is compared
+ * @returns {Array}
+ */
+module.exports.uniqueBy = (arr, getValue) => {
+  let unique = []
```
dirkmc:
```suggestion
  return [...new Map(arr.map((i) => [getValue(i), i])).values()]
```

vasco-santos (approved):
LGTM

src/utils.js
jacobheun: nice, updated

dirkmc (approved):
LGTM 👍

PR #337, opened 2019-04-26
fix: dont blindly add observed addresses to our list

jacobheun:
Until we can properly validate the observed address our peer tells us about, we shouldn't blindly add it to our address list. Until we have better NAT management we cant reliably validate that we're adding an appropriate address for ourselves.

This is a "good citizen" fix. Since we will advertise these addresses to new peers, we should do our due diligence to make sure they're reasonably accurate.

### DIFF
--- a/src/connection/manager.js
+++ b/src/connection/manager.js
@@ -179,12 +179,7 @@ class ConnectionManager {
               })
             }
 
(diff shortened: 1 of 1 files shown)


vasco-santos (approved):
👍

PR #340, opened 2019-06-05
fix: clear blacklist for peer when connection is established

jacobheun:
This PR clears the blacklist for a peer when it established a muxed connection to the peer. 

Previously, if you tried to connect to a peer that was temporarily unavailable, it would get blacklisted for a short period. If that peer came back online and connected to you, the blacklist wasn't getting cleared. This would prevent any application code from creating new streams to that peer because the dialer sees the blacklist. Incoming connections will now clear the blacklist to correct this.

Adding the test I also noticed the dialer queue `.add` function was not exiting early when things were blacklisting, so I corrected that.


vasco-santos (approved):
LGTM!

PR #342, opened 2019-08-01
chore: update deps and fix lint

jacobheun:


jacobheun:
codecov drop is due to the .aegir.js file changes getting picked up

vasco-santos (approved):
LGTM!

